PR #8, opened 2015-05-29
Use godep to manage dependencies

achanda:
https://github.com/tools/godep


gnawux:
Hi @achanda, Good idea, but looks you committed your whole local workspace

achanda:
@gnawux I think that's how it works. Godep saves a snapshot of the deps in the `Godep` folder. This can be committed in git and distributed. Thus, even if the deps change and it breaks the app, we always have a local working copy in source control.

carmark:
Hello @achanda,

Since there is also a third-party library in the code, I will not merge your pull request.  However, your idea is really great to build this dependency.  I did the work based on your idea to use `godep` to make this work.

Thanks again.

Regards,
-Simon

PR #44, opened 2015-08-18
builder bug fixes

carmark:
* fix the `CreatePod` function arguments
* fix the create vm function on darwin
* set the builder ‘s `ForceRemove` flag to `true`


laijs:
code LGTM

PR #46, opened 2015-08-19
bug-container-list

feiskyer:
Fix container list slice bounds out of range error.


feiskyer:
Another commit has fixed this problem, so this one should be closed.

PR #55, opened 2015-08-25
return error if get user failed

gao-feng:
this will destroy vm immediately.
rearrange codes by the way.


carmark:
LGTM

PR #65, opened 2015-09-15
generator the container's name when the image's name can not be used as container name

carmark:
Now, `hyper` will use image's name as the container's name, but the container's name can only be `^/?[a-zA-Z0-9][a-zA-Z0-9_.-]+$`.  So it will generator a new container's name randomly when the image's name is invalid for the rule of container's name.

### Test Result

```
root@ubuntu:/home/lei/src/github.com/hyperhq/hyper# ./hyper run danielbodart/crazyfast.build
POD id is pod-oRxYYYiucq
root@ubuntu:/home/lei/src/github.com/hyperhq/hyper# ./hyper list container
Container ID                                                      Name                         POD ID    Status
cbd68f11ed4a878d106ad1462aecfbbe1dbbb6c35718c35e26313975e2c5c128  berserk_mcclintock   pod-oRxYYYiucq    failed
```

### DIFF
--- a/client/run.go
+++ b/client/run.go
@@ -7,11 +7,13 @@ import (
 	"io/ioutil"
 	"net/url"
 	"os"
+	"regexp"
 	"strconv"
 	"strings"
 	"time"
 
 	"github.com/hyperhq/hyper/engine"
+	"github.com/hyperhq/hyper/lib/docker/pkg/namesgenerator"
 	"github.com/hyperhq/hyper/lib/promise"
 	"github.com/hyperhq/hyper/utils"
 	"github.com/hyperhq/runv/hypervisor/pod"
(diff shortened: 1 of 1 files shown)


gnawux:
but what's the problem in the example? the slash?

I's the normal case a image name having a slash to separate the repo name and the image name. should we do a special process on slash?

carmark:
@gnawux 

Yeah, the slash is invalid character in a container's name.

I also did think of doing a special process for that slash, but the user may not care about that container's name if container's name is not specified.   So a valid and random container's name should be all right.

gnawux:
@carmark 

I don't think users care about it either. However, I think we should do things consistently. If we use image name as the container name prefix, then we should use `NNNNN` as name prefix of image `RRRR/NNNNN`. Otherwise, we should never use image name as the prefix for all cases.

carmark:
sounds reasonable, I will update it.

carmark:
@gnawux 

updated, could you please review it again?

gnawux:
looks good

PR #66, opened 2015-09-15
configure the version in code with configured value defined in configure.ac

gnawux:
define the version in build time


laijs:
LGTM

PR #69, opened 2015-09-16
autogen.sh: should use bash instead of sh

gnawux:
in debian derived distro (in about recent 7-10 years) including
pervasive accessible ubuntu , the default sh is `dash` instead of
`bash`, and provides much less features than `bash`.

if run `autogen.sh` on ubuntu, will cause

    $ ./autogen.sh
    ./autogen.sh: 8: ./autogen.sh: function: not found

We should use `bash` explicitly.

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 srcdir=`dirname $0`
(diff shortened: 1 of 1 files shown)


gnawux:
pr #72 replaced this one

PR #74, opened 2015-09-17
pr70 does not probe driver if it is specified in configuration file

gnawux:
#pr 70 only probe driver in case the driver is empty string.


gnawux:
@laijs @carmark pleas confirm this whether it is a bug of the code or my eyes.

carmark:
Yes, it is definitely a regression bug caused by pr#70.

Code LGTM

PR #80, opened 2015-09-19
Makefile: set the `GOPATH` to Godep + PATH contains src/PROJECT_PATH

gnawux:
Now no need to set GOPATH to build hyper any more


laijs:
What happen if the runv is in the different GOPATH?

PR #84, opened 2015-09-22
Init tty connection before container run

gnawux:
originally we start pod, then attach tty to it. Then we may miss
the beginning of the tty output. Even miss all the output if the
progress exit at once.

this patch include the hyper part and runv part, to enable
establishing the tty connection before starting pod. The result
will be like:

    gnawux@sonic$ ./hyper run busybox ps aux
    PID   USER     COMMAND
        1 root     /init
        2 root     ps aux
    Successfully attached to pod(pod-xhZAmXwXCV)
    End of CmdExec(), Waiting for hijack to finish.
    POD id is pod-xhZAmXwXCV

### DIFF
--- a/client/attach.go
+++ b/client/attach.go
@@ -49,6 +49,10 @@ func (cli *HyperClient) HyperCmdAttach(args ...string) error {
 	}
 	v.Set("tag", tag)
 
+	return cli.hijackRequest("attach", podName, tag, &v)
+}
+
+func (cli *HyperClient) hijackRequest(method, pod, tag string, v *url.Values) error {
 	var (
 		hijacked = make(chan io.Closer)
 		errCh    chan error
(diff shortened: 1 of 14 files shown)


gnawux:
this PR won't built before https://github.com/hyperhq/runv/pull/33 got merged.

gnawux:
I have updated the patch to version 2, here is the explanation of this PR and https://github.com/hyperhq/runv/pull/33.

## Background

### Start State Transition of VM in runV

There are three states are related with this PR:

- For a VM, it is start when we call `vm.Launch()`, and in 'Init' state.
- When `vm.StartPod()` is called, it get into `Starting` state, and it associated with a Pod in this state.
- When `hyperstart` confirm the Pod is running, it is in `Running` state.

In the previous version, `vm.Attach()` can only be called after vm is associate with a Pod, i.e. it can not be invoked during `Init` state. The result is---

We can only call `vm.Attach()` after having called `vm.StartPod()`. which means the container command does not connected with a tty when it is starting, and if the command is a one-shot command such as `ps`, we can not get the output.

## About this patch

### In runV

We enabled `attach` in `Init` state by add a buffer, which cache the tty connections before Pod is existing, and then plumb them when the Pod is initiating before the Pod starting.

### In Hyper Daemon

When start pod, if has TtyIO connected, it will call `vm.Attach()` before `vm.StartPod()`

### In Hyper CLI

Hyper CLI will initial hijack stream for pod/start request if there need setup a tty connection.

carmark:
You may need do `go fmt` on the changes' files.

gnawux:
@carmark ok, will update this

gnawux:
@carmark formatted

PR #91, opened 2015-09-24
allow bind either directory or file for vfs volume driver

gnawux:
then we can bind host `/etc/resolv.conf` to guest

this will help fix https://github.com/hyperhq/hyper/issues/90

this patch include an example for resolve.conf mapping


gao-feng:
LGTM

laijs:
LGTM
test good. but the cache in the vm need to be disabled.

PR #100, opened 2015-09-28
Improve list functionality

gnawux:
- do not show the auxiliary containers by default, only show if with `-x` option or
  `auxiliary=yes` exist in request URL parameter
- add `-p` option or `pod={pod_id}` in request URL parameter to list only the specified
  pod, rather than list all.

with this patch, exec {pod-id} will not exec to the service discovery auxiliary container


feiskyer:
lgtm

PR #102, opened 2015-09-28
cli: return non-zero value when command return error

### LINKED ISSUE #99: hyper pull always returns exit code 0 even if it errors in some way
If you run 

```
hyper pull someinvalidimagename
```

It will return 0

gnawux:
this should fix #99


gao-feng:
LGTM

PR #107, opened 2015-10-03
add volumes in pod if images contains volumes configuration

gnawux:
such as `/var/cache/nginx` in official `nginx` image


feiskyer:
LGTM

PR #110, opened 2015-10-12
use run "-i" or "tty" field to specify whether it is interactive

gnawux:
like shells

together with runv https://github.com/hyperhq/runv/pull/48


gnawux:
o... it is failed as we merged hyperhq/runv#49 , more dependency are included

PR #116, opened 2015-10-24
not only init volumeInfo struct when not creating volume

gnawux:
it need fill other fields such as name, filepath, fstype etc.
do different operation for different volume types


gnawux:
tested with the following PODs:

- `examples/service-discovery.pod`: vol created by image, vol pass through virtfs
- `examples/file-mapping.pod`: file mapping through virtfs
- `examples/insert-file.pod`: insert file contents
- `examples/with-volume.pod`: multiple vols mounted by multiple containers.

PR #117, opened 2015-10-25
Close logger when pod stopped

gao-feng:


gnawux:
I think it's also need to close logger if startpod failed.

gao-feng:
updated

gnawux:
should change line 706 `err :=` to `err =` in order to guarantee the defer behavior.

PR #132, opened 2015-11-16
run: launch vm before create pod

gnawux:
so that we can make create+prepare parallel run with vm launch.
This intents to fix the performance regression introduced by the code formalization

### DIFF
--- a/client/vm.go
+++ b/client/vm.go
@@ -2,6 +2,8 @@ package client
 
 import (
 	"fmt"
+	"net/url"
+	"strconv"
 	"strings"
 
 	"github.com/hyperhq/hyper/engine"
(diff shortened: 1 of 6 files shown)


carmark:
```
+	var (
+		spec  pod.UserPod
+		async = true
+	)
+	json.Unmarshal([]byte(podJson), &spec)
+
+	vmId, err := cli.CreateVm(spec.Resource.Vcpu, spec.Resource.Memory, async)
+
 	podId, err := cli.CreatePod(podJson, opts.Remove)
```

This process still starts pod one by one, creating vm and pod, then start it.  How about using API interface of `/pod/run`?  It will create pod during vm's creating and starting, and then associate the pod into the vm.

gnawux:
@carmark no, the async flag should not wait the vm started if it's works well, and if there is some problem in async creating vm, I will fix it. I don't want to keep the pod/run api, which is confused

client/vm.go
```diff
@@ -22,36 +24,62 @@ func (cli *HyperClient) HyperCmdVm(args ...string) error {
...
+	return nil
+}
+
+func (cli *HyperClient) CreateVm(cpu, mem int, async bool) (id string, err error) {
+	id = ""
+	err = nil
+	var (
+		body       []byte
+		remoteInfo *engine.Env
+	)
```
carmark: * `err = nil` seems useless
* how about moving `id=""` into `var ()`?

carmark:
OK, fine for me.  I think it is good to remove those code(`/pod/run`) with this pr.

gnawux:
@carmark I tested this PR, and the /pod/run, and the current master in one box, can not find significant performance difference, will test more cases before merge

gnawux:
testd with

- master
- this pr
- previous one shot run api

the result in short ---

- for `overlay`, almost the same boot time for the three method.
- for `devicemapper`, this pr faster about 1/4 than the other two.

will do clean up and merge this pr.

PR #147, opened 2015-12-10
fix fail to start pod

gao-feng:
if pod is stopped automatically, we should reset pod otherwise
start pod will fail with error:
./hyper ERROR: Error from daemon's response: pod pod-JfRVQkHSZc is already running


feiskyer:
LGTM

PR #151, opened 2015-12-28
add libvirt check in autoconf

gao-feng:

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -72,6 +72,19 @@ AC_TYPE_UINT8_T
 AC_FUNC_MALLOC
 AC_CHECK_FUNCS([strdup])
 
+LIBVIRT_REQUIRED="1.2.2"
+
+AC_ARG_WITH([libvirt],
+            [AS_HELP_STRING([--without-libvirt],
+                            [run hyperd with libvirt])],
+            [with_libvirt=no],[with_libvirt=yes])
+
+if test "$with_libvirt" = yes; then
+	PKG_CHECK_MODULES([libvirt], [libvirt >= $LIBVIRT_REQUIRED], [], [with_libvirt=no])
+fi
+
+AM_CONDITIONAL([WITH_LIBVIRT], [test "x$with_libvirt" == "xyes"])
+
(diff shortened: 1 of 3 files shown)


gnawux:
will it compile fail if no libvirt present? If so, we need some code to disable libvirt related code if libvirt develop files are not available.

gao-feng:
--without-libvirt, same with xen

configure.ac
```diff
@@ -72,6 +72,20 @@ AC_TYPE_UINT8_T
...
+LIBVIRT_REQUIRED="1.2.2"
+
+AC_ARG_WITH([libvirt],
+            [AS_HELP_STRING([--without-libvirt],
+                            [run hyperd with libvirt])],
+            [with_libvirt=no],[with_libvirt=yes])
+
+if test "$with_libvirt" = yes; then
+	PKG_CHECK_MODULES([libvirt], [libvirt >= $LIBVIRT_REQUIRED], [], [
+		AC_MSG_ERROR([libvirt-devel >= $LIBVIRT_REQUIRED is required or reconfigure runv with --without-libvirt])])
```
laijs: "reconfigure runv with", it should be hyper.

laijs: it should not fail when there is no libvirt-dev.

laijs: --with-xen will be silent turn off when there is no xen-dev installed. I think it should complain and fail.

--with-xen & xen-dev  => xen
--with-xen & no xen-dev  =>  fail
--without-xen & xen-dev  => no xen
--without-xen & no xen-dev  => no xen
xen-dev  => xen
no xen-dev  => no xen

the same for libvirt

gao-feng: it won't fail, if there is no libvirt-dev or the version is incorrect, you should reconfigure hyper/runv with --without-libvirt, please check https://github.com/hyperhq/runv/pull/96.

We want xen&libvirt driver being available by default, so we outthrow warning if there is no xen or libvirt available,and user can use --without- option to reconfigure hyper/runv.

PR #153, opened 2016-01-05
set the deafult value for the flags

laijs:
recover the original behavior before 0420389d21b0d9630bec23904ae37d730be77b16


feiskyer:
LGTM

Issue #134, opened 2015-11-19
Add labels support to hyper pod


feiskyer:
Add labels support to hyper. This feature is essential for kubelet and cAdvisor integration.

Such as:

```sh
# create pod with labels
hyper run --label com.example.group="web servers" \
   --label com.example.environment="production" \
   busybox top

# update labels
hyper label pod-abcd com.example.group="frontend"
```

feiskyer:
Merged as #156

PR #156, opened 2016-01-06
Add labels for hyper pod

feiskyer:
Implements #134 

Requires https://github.com/hyperhq/runv/pull/98

Notes: travis will be erroring util https://github.com/hyperhq/runv/pull/98 got merged.

```sh
$ hyper run --label a=b --label c=d nginx
POD id is pod-IdlABgYRDx

$ curl hyper-api/pod/info?podName=pod-IdlABgYRDx | jq .spec.labels
{
  "a": "b",
  "c": "d"
}
```

### DIFF
--- a/client/run.go
+++ b/client/run.go
@@ -197,11 +197,11 @@ func (cli *HyperClient) JsonFromCmdline(cmdArgs, cmdEnvs, cmdPortmaps []string, 
 	}
 
 	for _, v := range cmdLabels {
-		eql := strings.Index(v, "=")
-		if eql > 0 {
-			labels[v[:eql]] = v[eql+1:]
+		label := strings.Split(v, "=")
+		if len(label) == 2 {
+			labels[label[0]] = label[1]
 		} else {
-			labels[v] = ""
+			return "", fmt.Errorf("Label '%s' is not in 'k=v' format", v)
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


daemon/pod.go
```diff
@@ -57,6 +58,78 @@ func (daemon *Daemon) CmdPodCreate(job *engine.Job) error {
...
+
+	for k := range labels {
+		if _, ok := pod.spec.Labels[k]; ok && !override {
+			return fmt.Errorf("Can't update label %s without override", k)
+		}
+	}
+
+	for k, v := range labels {
+		pod.spec.Labels[k] = v
+	}
```
carmark: how about combing the code from line 102 to 110?

client/run.go
```diff
@@ -194,6 +196,15 @@ func (cli *HyperClient) JsonFromCmdline(cmdArgs, cmdEnvs, cmdPortmaps []string,
...
 
+	for _, v := range cmdLabels {
+		eql := strings.Index(v, "=")
+		if eql > 0 {
+			labels[v[:eql]] = v[eql+1:]
+		} else {
+			labels[v] = ""
+		}
+	}
+
```
carmark: I think we should not support the label format which does not contain "=".  And it will be better to delete the space if it has via `strings.TrimSpace(v)`.

feiskyer: Updated. And shell will delete the trailing space so TrimSpace is not needed.

carmark:
LGTM

PR #159, opened 2016-01-07
Add cpu/memory to pod info api

feiskyer:
Add cpu/memory to pod info api, so we can check cpu and memory info of pod spec.

```sh
[root@host ~]# curl -s 'hyper-api/pod/info?podName=pod-pNBQIpaHzp'  | jq .spec.memory
128
[root@host ~]# curl -s 'hyper-api/pod/info?podName=pod-pNBQIpaHzp'  | jq .spec.vcpu
1
```


carmark:
LGTM

PR #161, opened 2016-01-07
do not specify kvm hypervisor driver in auto test

gao-feng:
let hyper to specify driver automatically.

### DIFF
--- a/hack/test-cmd.sh
+++ b/hack/test-cmd.sh
@@ -86,8 +86,8 @@ runTests() {
   execdriver="$1"
   stordriver="$2"
   echo "Testing hyperd with execdriver: $1, storage driver: $2"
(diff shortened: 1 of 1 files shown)


hack/test-cmd.sh
```diff
@@ -126,7 +126,7 @@ hyper_storage_drivers=(
 )
 
 hyper_exec_drivers=(
-  "kvm"
+  ""
 )
```
carmark: how about adding `qemu` and `libvirt`?

gao-feng: driverloader will decide the hypervisor. In the trivals environment, only qemu is available, so actually we test both driverloader and qemu driver.

carmark:
LGTM

PR #165, opened 2016-01-10
let the test-cmd.sh could run locally

gnawux:
$ sudo HYPER_RUNTIME=/var/lib/hyper hack/test-cmd.sh

and make sure the stop can kill the hyperd process


carmark:
LGTM

PR #170, opened 2016-01-11
prefetching workflow

laijs:
prefetch vm when create pod

use prefetching workflow to run/start pods
implement dummy prefetchVm() (simply calls StartVm())
reduce one step for `hyper run`

move ProcessPodBytes() out from podlist lock


gnawux:
looks #132 was reverted. Why not just "hide" prefetchVm in createVm?

laijs:
it changed the api. a new design without changing any api will be ready. close it.

PR #183, opened 2016-01-17
remove the non-container output for attach and attached run

gnawux:
make it easier to process the output of hyper run


carmark:
LGTM

PR #184, opened 2016-01-18
Fix rbd volume for libvirt driver

feiskyer:
Fix https://github.com/hyperhq/runv/pull/109 https://github.com/hyperhq/runv/pull/108

Depends on https://github.com/hyperhq/runv/pull/110

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -17,6 +17,11 @@
 			"Rev": "5cfc02f2789e6e98d8546e40b29d907aedc4c590"
 		},
 		{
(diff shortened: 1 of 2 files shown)


gnawux:
should not modify the code under Godep directly. code here will be overwritten when we update the dependencies.

feiskyer:
Here is indeed a dependency update. We can't wait for libvirt-go upstream to accept it.

laijs:
then you should also update the Godeps.json

feiskyer:
sure

PR #188, opened 2016-01-22
support Pause container for hyper

gao-feng:
based on https://github.com/hyperhq/runv/pull/111

### DIFF
--- a/client/pause.go
+++ b/client/pause.go
@@ -1,0 +1,45 @@
+package client
+
+import (
+	"fmt"
+	"net/url"
+	"strings"
+
+	"github.com/hyperhq/hyper/engine"
+	gflag "github.com/jessevdk/go-flags"
+)
+
+func (cli *HyperClient) HyperCmdPause(args ...string) error {
+	var parser = gflag.NewParser(nil, gflag.Default|gflag.IgnoreUnknown)
+	parser.Usage = "pause Pod\n\nPause the pod"
+	args, err := parser.ParseArgs(args)
+	if err != nil {
+		if !strings.Contains(err.Error(), "Usage") {
+			return err
+		} else {
+			return nil
+		}
+	}
+	if len(args) == 0 {
+		return fmt.Errorf("Can not accept the 'pause' command without Pod ID!")
+	}
+
+	v := url.Values{}
+	v.Set("podId", args[0])
+
+	body, _, err := readBody(cli.call("POST", "/pod/pause?"+v.Encode(), nil, nil))
+	if err != nil {
+		return err
+	}
+
+	out := engine.NewOutput()
+	if _, err = out.AddEnv(); err != nil {
+		return err
+	}
+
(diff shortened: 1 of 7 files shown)


client/pause.go
```diff
@@ -0,0 +1,44 @@
...
+		} else {
+			return nil
+		}
+	}
+	if len(args) == 1 {
+		return fmt.Errorf("Can not accept the 'pause' command without Container ID!")
+	}
+
+	v := url.Values{}
+	v.Set("container", args[1])
```
carmark: should we check the args number to avoid the array out of range?

client/unpause.go
```diff
@@ -0,0 +1,44 @@
...
+		} else {
+			return nil
+		}
+	}
+	if len(args) == 1 {
+		return fmt.Errorf("Can not accept the 'unpause' command without Container ID!")
+	}
+
+	v := url.Values{}
+	v.Set("container", args[1])
```
carmark: same as the above

server/server.go
```diff
@@ -594,6 +594,40 @@ func postContainerRename(eng *engine.Engine, version version.Version, w http.Res
...
+
+func postContainerUnpause(eng *engine.Engine, version version.Version, w http.ResponseWriter, r *http.Request, vars map[string]string) error {
+	if err := parseForm(r); err != nil {
+		return err
+	}
+
+	job := eng.Job("unpause", r.Form.Get("container"))
+	stdoutBuf := bytes.NewBuffer(nil)
+
+	job.Stdout.Add(stdoutBuf)
```
carmark: look like this `stdoutBuf` is useless, should we delete it?

client/pause.go
gao-feng: Already make sure args array at least has one arg. Doesn't line 23 did this job? or we need another check here?

carmark:
Could you please move the `"github.com/hyperhq/hyper/engine"` into the second part of imported packages?

server/server.go
```diff
@@ -1175,23 +1203,25 @@ func createRouter(eng *engine.Engine, logging, enableCors bool, corsHeaders stri
...
-			"/service/update":   postServiceUpdate,
-			"/tty/resize":       postTtyResize,
-			"/vm/create":        postVmCreate,
+			"/auth":              postAuth,
+			"/attach":            postAttach,
+			"/container/create":  postContainerCreate,
+			"/container/commit":  postContainerCommit,
+			"/container/rename":  postContainerRename,
+			"/container/pause":   postContainerPause,
+			"/container/unpause": postContainerUnpause,
```
laijs: I think we should provide /pod/[un]pause, since we do it under the hood.
we may provide "/container/pause" in future when we can pause it individually

gao-feng: @laijs pause api introduced here is for docker commit container, commit need to pause container. we provide this api to consist with docker api. the internal implements can be change when we can really implement pause container.

PR #195, opened 2016-01-30
docker will reject to create container if the image do not have default commands

gnawux:
for example, fedora image

we should send command array to docker code if there are some.


daemon/pod.go
```diff
@@ -310,8 +310,13 @@ func (p *Pod) InitContainers(daemon *Daemon, dclient DockerInterface) (err error
 
 		glog.V(1).Info("trying to init container ", c.Name)
 
+		cmds := c.Entrypoint
+		if len(c.Command) > 0 {
+			cmds = append(cmds, c.Command...)
```
laijs: cmds = append(append([]string{}, cmds...), c.Command...)

PR #197, opened 2016-01-31
Fix command line argument process

gnawux:
There are several bugs/issue in client command line argument processing, for example

    hyper run -t

or

    hyper build -t "myrepo/myimg"

the above incorrect command line will lead to client panic.

The other problem details include:

## Parse or ParseArgs

In short `Parse()` process os.Args[1:] and `ParseArgs(args []string)` process `args`

The commands have `args` in command line, such as

    func (cli *HyperClient) HyperCmdBuild(args ...string) error

when user types ` hyper build -t "myrepo/myimg"`

- the `args` will be `-t`, `myrepo/myimg`
- the `os.Args` should be `hyper`, `build`, `-t`, `myrepo/myimg`

if we use

    args, err := parser.Parse()

the args will be `build`, and if we check if it is empty, it will never be empty,
because it always has a `build` here even if we forget to provide the path

And if we use

    args, err := parser.ParseArgs(args)

We will get an empty array, through which we can find the input error.

As we have already provide args as the method parameter, we should use
`ParseArgs`

## Should not check args len at the very beginning

we checked `len(args)` some of commands before argument processing,
for example, the `run` checked.

However, if we input `hyper run -t`, it won't find the problem because it has
a `-t`, the problems will be found by following parameters check. The pre-processing
check is trivial.

### DIFF
--- a/client/attach.go
+++ b/client/attach.go
@@ -14,7 +14,7 @@ import (
 func (cli *HyperClient) HyperCmdAttach(args ...string) error {
 	var parser = gflag.NewParser(nil, gflag.Default)
 	parser.Usage = "attach CONTAINER\n\nAttach to the tty of a specified container in a pod"
-	args, err := parser.Parse()
+	args, err := parser.ParseArgs(args)
 	if err != nil {
 		if !strings.Contains(err.Error(), "Usage") {
 			return err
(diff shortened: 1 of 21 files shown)


laijs:
LGTM

PR #202, opened 2016-02-03
test case for hostname field in pod spec

gnawux:
hyperhq/runv#117 added dedicate hostname field in pod, this is the related test case


feiskyer:
LGTM

PR #218, opened 2016-02-25
exit event handler when receiving POD_STOPPED event

gao-feng:
based on https://github.com/hyperhq/runv/pull/123

in replace, the old pod is stopped, but the event handler is still
running and operating the vm, after new pod start, there will be two
event handlers to operate the same vm, this may cause panic on close
vm's fanout chan twice. and will cause two handler to operate the same
podstatus when we start the old pod again. such as

hyper start pod-a
hyper replace -o pod-a -n pod-b
hyper start pod-a
hyper stop pod-b

you will see the status of pod-a is failed too. but the hypervisor is still
running.

fix this problem by finishing the event handler of old pod when receiving
the pod stopped event(in hyper) and don't close the fanout chan is this
case(vm is still running). this will make sure only one handler is using
the vm.

### DIFF
--- a/daemon/pod.go
+++ b/daemon/pod.go
@@ -916,22 +916,22 @@ func (daemon *Daemon) RestartPod(mypod *hypervisor.PodStatus) error {
 }
 
 func hyperHandlePodEvent(vmResponse *types.VmResponse, data interface{},
-	mypod *hypervisor.PodStatus, vm *hypervisor.Vm) bool {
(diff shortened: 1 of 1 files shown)


bergwolf:
LGTM!

gao-feng:
close,we decide to disable the features running pods on same vm, it makes things complicated

gao-feng:
close,we decide to disable the features running pods on same vm, it makes things complicated

PR #225, opened 2016-03-08
respect the Tty option

laijs:


gnawux:
restarted tests, and passed

PR #226, opened 2016-03-10
hyper client cmd line option cleanup

bergwolf:
clean up hyper client command line options a bit:
1. hyper rmi -f defaults to false now, because go-flags does not work well with boolean option defaulting to true, in which case false cannot be specified by users. Besides, hyper does support non-force rmi by rejecting it if there is a pos using the image.
2. remove redundant value-name in bool options. go-flags only use it to generate help message and manpage, where bool option would print its own default value anyway.


gao-feng:
LGTM

PR #231, opened 2016-03-17
update Docker daemon configuration

gao-feng:
enable debug mode
disable iptables rules
disable nat MASQUERADE rules


gnawux:
LGTM

PR #232, opened 2016-03-18
get tty info from container/info interface

gnawux:
- support query with container id, prefix, and name (exact id is prefered)
- if there are more than one container share the specified prefix, will return not error

test results:

    [root@h8s-single hyper]# curl http://localhost:9999/v0.5/container/info?container=d29a848c0d58d2660b52d05c6aca2a53d9d464c791e97323c86be7baa3680d3c
    {"name":"/busybox-2919434435","containerID":"d29a848c0d58d2660b52d05c6aca2a53d9d464c791e97323c86be7baa3680d3c","podID":"pod-YPEragSJvS","image":"busybox","imageID":"sha256:3240943c9ea3f72db51bea0a2428e83f3c5fa1312e19af017d026f9bcf70f84b","commands":[],"args":[],"workingDir":"/","ports":[],"env":[],"volumeMounts":[],"tty":false,"imagePullPolicy":"","status":{"name":"busybox-2919434435","containerID":"d29a848c0d58d2660b52d05c6aca2a53d9d464c791e97323c86be7baa3680d3c","phase":"pending","waiting":{"reason":"Pending"},"running":{"startedAt":""},"terminated":{"exitCode":0,"reason":"","message":"","startedAt":"","finishedAt":""}}}
    [root@h8s-single hyper]# curl http://localhost:9999/v0.5/container/info?container=d29a848c0d58d2660b52d05c6aca2a53d9d46
    {"name":"/busybox-2919434435","containerID":"d29a848c0d58d2660b52d05c6aca2a53d9d464c791e97323c86be7baa3680d3c","podID":"pod-YPEragSJvS","image":"busybox","imageID":"sha256:3240943c9ea3f72db51bea0a2428e83f3c5fa1312e19af017d026f9bcf70f84b","commands":[],"args":[],"workingDir":"/","ports":[],"env":[],"volumeMounts":[],"tty":false,"imagePullPolicy":"","status":{"name":"/busybox-2919434435","containerID":"d29a848c0d58d2660b52d05c6aca2a53d9d464c791e97323c86be7baa3680d3c","phase":"pending","waiting":{"reason":"Pending"},"running":{"startedAt":""},"terminated":{"exitCode":0,"reason":"","message":"","startedAt":"","finishedAt":""}}}
    [root@h8s-single hyper]# curl http://localhost:9999/v0.5/container/info?container=busybox-9514458980
[...]

PR #233, opened 2016-03-18
update pod info and container info

gnawux:
- use unify structure for pod and container info
- update cmd and args, as now it comes from inspect
- use image in spec instead of status, as the new status store image id instead of image name in this field

and keeps the /container/info results exactly same as the result in #232


feiskyer:
LGTM

PR #235, opened 2016-03-18
remove bug sendStdin

laijs:
sendStdin is useless, we don't wait for the stdin,
we only wait for stdout.

sendStdin is bug, when stdin&stdout closed at the same
time, we will blocked forever on sendStdin


gnawux:
Good

PR #241, opened 2016-03-29
Adjust the locking mechanism in daemon

gnawux:
- pod list lock is dedicated for pod list internal access now
- set pod return status success only if the pod return with all 0 value, i.e. set failed if user stop it. and some other stop processing adjustment
- move the demon methods about create/start pod to a new file
- move all leveldb code to daemondb package, and remove unnecessary daemon method for those have in daemondb
- remove some old flag or methods, including autoremove, GetPod/RunPod

### DIFF
--- a/daemon/podlist.go
+++ b/daemon/podlist.go
@@ -61,7 +61,7 @@ func (pl *PodList) Delete(id string) {
 
 func (pl *PodList) GetByName(name string) *Pod {
 	pl.mu.RLock()
-	defer pl.mu.Unlock()
+	defer pl.mu.RUnlock()
 
 	return pl.findUnsafe(func(p *Pod) bool {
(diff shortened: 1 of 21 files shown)


daemon/podlist.go
```diff
@@ -101,7 +101,7 @@ func (pl *PodList) GetByContainerId(cid string) (*Pod, bool) {
 
 func (pl *PodList) GetByContainerIdOrName(cid string) (*Pod, int, bool) {
 	pl.mu.RLock()
-	defer pl.mu.Unlock()
+	defer pl.mu.RUnlock()
```
gao-feng: I saw GetByName GetByContainerId have the same problem.

gao-feng:
LGTM

PR #243, opened 2016-03-29
only check pod ip if pod is running

gnawux:
and improve the message in case start failed


gao-feng:
LGTM

PR #254, opened 2016-04-03
store hosts file in tmpfs

gao-feng:
The deault size of tmpfs is 1M


laijs:
LGTM

PR #262, opened 2016-04-09
for non-tty container, write log should use stdcopy

gnawux:
This is the behavior compatible to docker(tm). However, because hyperctl
always try using stream as tty terminal. Need further modification on hyperctl.
should we do the further improvement in another PR?


laijs:
LGTM

PR #274, opened 2016-04-13
 remove DockerVolume from pod spec

gao-feng:
remove DockerVolume from pod spec

based on https://github.com/hyperhq/runv/pull/176

### DIFF
--- a/daemon/pod.go
+++ b/daemon/pod.go
@@ -390,7 +390,7 @@ func (p *Pod) parseContainerJsons(daemon *Daemon, jsons []*dockertypes.Container
 			}
 		}
 
-		processImageVolumes(info, info.ID, p.spec, &p.spec.Containers[i], ci.Initialize)
+		p.processImageVolumes(info, info.ID, &p.spec.Containers[i])
 
 		p.ctnStartInfo = append(p.ctnStartInfo, ci)
 		glog.V(1).Infof("Container Info is \n%v", ci)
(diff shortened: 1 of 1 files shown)


daemon/pod.go
```diff
@@ -536,7 +537,7 @@ func (p *Pod) setupServices() error {
 	return err
 }
 
-// PrepareEtcHosts sets /etc/hosts for each container
+// Setu[EtcHosts sets /etc/hosts for each container
 func (p *Pod) setupEtcHosts() (err error) {
```
laijs: setup

daemon/pod.go
```diff
@@ -728,35 +729,66 @@ func (p *Pod) setupMountsAndFiles(sd Storage) (err error) {
...
-		p.volumes = append(p.volumes, vol)
+		if vol, ok := p.volumes[v.Name]; ok {
+			vol.Filepath = volInfo.Filepath
+			vol.Fstype = volInfo.Fstype
+			vol.Format = volInfo.Format
+			continue
+		}
+
+		p.volumes[v.Name] = volInfo
+	}
```
laijs: I think only `p.volumes[v.Name] = volInfo` is Ok,  we don't need to copy it one field by field if the volume existed.

laijs: It copies and keep DockerVolume. ignore my comments.

daemon/storage.go
```diff
@@ -40,6 +40,7 @@ type Storage interface {
 
 	PrepareContainer(id, sharedir string) (*hypervisor.ContainerInfo, error)
 	InjectFile(src io.Reader, containerId, target, rootDir string, perm, uid, gid int) error
+	// CreateVolume should check if volume is existed to avoid allocating volume mulit time
 	CreateVolume(daemon *Daemon, podId, shortName string) (*hypervisor.VolumeInfo, error)
```
laijs: does all the store drivers obeyed it?

daemon/pod.go
```diff
@@ -422,7 +418,8 @@ func (p *Pod) createVolumes(daemon *Daemon) error {
 
 	sd := daemon.Storage
 	for i := range p.spec.Volumes {
-		if p.spec.Volumes[i].Source == "" {
+		// hyperd doesn't 'remove' vfs volume
+		if p.spec.Volumes[i].Driver != "vfs" {
 			vol, err = sd.CreateVolume(daemon, p.id, p.spec.Volumes[i].Name)
```
gnawux: I think this patch changed the semantics of the `CreateVolume`. In the previous code, the daemon check the user spec, and only create volume when the user doesn't specify the source of the volume.

Issue #278, opened 2016-04-14
Revisit hyper client and api


feiskyer:
The hyper client code should be separated from hyper CLI. And we should also add an optional gRPC api to improve performance.

- [x] Separate client code and api code #263
- [ ] Prepare types for both current api and gRPC
- [ ] Clean up `client/api/` code, move CLI related function to CLI itself
- [ ] Add optional gPRC api

xlgao-zju:
I will take `GET, /info` for practice.

resouer:
`stats`, `labels` will be send after `start`

heartlock:
I will take POST, /image/load   POST, /auth    POST, /image/build

gnawux:
@feiskyer Do we still need this issue for 0.8 and later release?

feiskyer:
There are still some interfaces missing, e.g. image build, container commit and so on. And current http api is still not respect of restful.

PR #279, opened 2016-04-14
Prepare types for both current api and gRPC

feiskyer:
Prepare types for both current api and gRPC, using types generated by proto.

Part of #278


gnawux:
gRPC is welcome

types/types.proto
```diff
@@ -0,0 +1,127 @@
...
+message ContainerInfo {
+  string name                       = 1;
+  string containerID                = 2;
+  string image                      = 3;
+  string imageID                    = 4;
+  repeated string commands          = 5;
+  repeated string args              = 6;
+  string workingDir                 = 7;
+  repeated ContainerPort ports      = 8;
+  repeated EnvironmentVar env       = 9;
```
laijs: `map<string,string> env` might be a choice

types/types.proto
```diff
@@ -0,0 +1,127 @@
...
+
+// TODO: embed Container message to avoid repetition
+message ContainerInfo {
+  string name                       = 1;
+  string containerID                = 2;
+  string image                      = 3;
+  string imageID                    = 4;
+  repeated string commands          = 5;
+  repeated string args              = 6;
+  string workingDir                 = 7;
```
laijs: workdir

types/types.proto
```diff
@@ -0,0 +1,127 @@
...
+}
+
+// TODO: embed Container message to avoid repetition
+message ContainerInfo {
+  string name                       = 1;
+  string containerID                = 2;
+  string image                      = 3;
+  string imageID                    = 4;
+  repeated string commands          = 5;
+  repeated string args              = 6;
```
laijs: why it has both `commands` and  `args`

types/types.proto
```diff
@@ -0,0 +1,127 @@
...
+    RunningStatus running   = 5;
+    TermStatus terminated   = 6;
+}
+
+// TODO: embed Container message to avoid repetition
+message ContainerInfo {
+  string name                       = 1;
+  string containerID                = 2;
+  string image                      = 3;
+  string imageID                    = 4;
```
laijs: why it has both image and imageID

types/types.proto
```diff
@@ -0,0 +1,127 @@
...
+
+message EnvironmentVar {
+    string env    = 1;
+    string value  = 2;
+}
+
+message VolumeMount {
+    string name       = 1;
+    string mountPath  = 2;
+    bool readOnly     = 3;
```
laijs: use  `repeated string option` instead
the userpod.json will also be updated.

types/types.proto
feiskyer: Same meaning with entrypoint and commands

types/types.proto
```diff
@@ -0,0 +1,127 @@
...
+  repeated string args              = 6;
+  string workingDir                 = 7;
+  repeated ContainerPort ports      = 8;
+  repeated EnvironmentVar env       = 9;
+  repeated VolumeMount volumeMounts = 10;
+  bool tty                          = 11;
+  string imagePullPolicy            = 12;
+	string podID                      = 13;
+	ContainerStatus status            = 14;
+}
```
laijs: repeated processinfo processes

types/types.proto
feiskyer: why `repeated string option`? A struct is more clear than repeated string.

laijs: options a complicated, A struct is overkill.
see also https://github.com/opencontainers/runtime-spec/blob/master/specs-go/config.go#L93-L94

feiskyer: Thanks. Another ref https://github.com/opencontainers/runc/blob/master/libcontainer/specconv/spec_linux.go#L583-L608

types/types.proto
feiskyer: image is actually the name of image

types/types.proto
feiskyer: Where is processinfo defined?

PR #282, opened 2016-04-14
Fix Build problem

gao-feng:


gnawux:
LGTM, the #289 is not introduced by this PR

PR #288, opened 2016-04-15
rename to hyperd

gnawux:
update the travis icon from hyper to hyperd, and update some detail in readme


gnawux:
to pass the test, we need to update all the `import` in project, should we merge #282 before this? However, the #282 won't pass too if I do not modify the imports

Issue #286, opened 2016-04-15
Get ExitTime for hyper pod


feiskyer:
Get ExitTime for hyper pod while pod becoming success or failed.

This is required to determine when the pod is exited.

gnawux:
Do you mean got it from `info` API?

feiskyer:
Y, in `/pod/info` api.

feiskyer:
Implemented in #295.

PR #295, opened 2016-04-19
Add finish time for stopped Pod

feiskyer:
Implement #286


daemon/pod.go
```diff
@@ -1004,7 +1004,8 @@ func hyperHandlePodEvent(vmResponse *types.VmResponse, data interface{},
 		if mypod.Status == types.S_POD_RUNNING { // not received finished pod before
 			stopLogger(mypod)
 			mypod.Status = types.S_POD_FAILED
-			mypod.SetContainerStatus(types.S_POD_FAILED)
+			mypod.FinishedAt = time.Now().Format("2006-01-02T15:04:05Z")
+			mypod.SetPodContainerStatus(types.S_POD_FAILED)
 		}
```
gnawux: https://github.com/hyperhq/runv/blob/master/hypervisor/pod.go#L152

within `SetPodContainerStatus`, the `mypod.status` and `mypod.FinishedAt` are both set. This is why I said here is a bug

gnawux:
ok, this patch should work well. And let's talk about merge `SetContainerStatus` and `setPodContainerStatus` later, which is an independent topic.

PR #301, opened 2016-04-21
rm: if vm in pod is still present, wait it stop before remove resource

gnawux:
- use lock protect unset pod.vm
- when delete, if vm is present, set pod status to be None
- when vm stopped, if status is None, reclaim the resource
- when list, do not display the removed pods

### DIFF
--- a/daemon/rm.go
+++ b/daemon/rm.go
@@ -36,19 +36,28 @@ func (daemon *Daemon) CleanPod(podId string) (int, string, error) {
 		}
 	}
 
-	os.RemoveAll(path.Join(utils.HYPER_ROOT, "services", podId))
-	pod.cleanupEtcHosts()
-	os.RemoveAll(path.Join(utils.HYPER_ROOT, "hosts", podId))
-
-	daemon.db.DeletePod(podId)
-	daemon.RemovePod(podId)
-	if pod.status.Type != "kubernetes" {
-		daemon.RemovePodContainer(pod)
+	pod.Lock()
+	defer pod.Unlock()
+	if pod.vm != nil {
+		pod.status.Status = types.S_POD_NONE
+		return code, cause, err
 	}
-	daemon.DeleteVolumeId(podId)
-	code = types.E_OK
 
-	return code, cause, nil
+	return daemon.RemovePodResource(pod)
+}
+
+func (daemon *Daemon) RemovePodResource(p *Pod) (int, string, error) {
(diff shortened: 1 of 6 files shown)


daemon/rm.go
```diff
@@ -36,19 +36,30 @@ func (daemon *Daemon) CleanPod(podId string) (int, string, error) {
...
-	daemon.DeleteVolumeId(podId)
-	code = types.E_OK
 
-	return code, cause, nil
+	return daemon.RemovePodResource(pod)
+}
+
+func (daemon *Daemon) RemovePodResource(p *Pod) (int, string, error) {
+	os.RemoveAll(path.Join(utils.HYPER_ROOT, "services", p.id))
+	os.RemoveAll(path.Join(utils.HYPER_ROOT, "hosts", p.id))
```
feiskyer: Is pod.cleanupEtcHosts() needed here?

gnawux: it is called when pod stopped.

gnawux: https://github.com/hyperhq/hyperd/pull/301/files#diff-cfb48529ee9117af9bd2e69c80e66291R25

every time pod stopped, this method will be called. No matter whether remove the pod.

feiskyer: Y, thanks.

daemon/rm.go
```diff
@@ -36,19 +36,30 @@ func (daemon *Daemon) CleanPod(podId string) (int, string, error) {
...
+
+	daemon.db.DeletePod(p.id)
+	daemon.RemovePod(p.id)
+	if p.status.Type != "kubernetes" {
+		daemon.RemovePodContainer(p)
+	}
+	daemon.DeleteVolumeId(p.id)
+	code := types.E_OK
+
+	return code, "", nil
```
gao-feng: return E_OK here

gnawux: ok, this is just copied from the original code.  

will update this, and are there any other comments.

gao-feng:
seems like StopPod(WithinLock) holds the lock of pod and then set the p.vm to nil first, then the PodStopped will not cleanup resource, should we stop setting p.vm to nil in StopPod, leave this job to PodStopped.

gnawux:
@gao-feng, I agree with you, set `p.vm` to `nil` only in `podStopped`. will update this pr in couple of minutes.

gao-feng:
LGTM

PR #304, opened 2016-04-25
rm: remove pod from podlist at last

gnawux:
- to run pod and rm, should let it attach, otherwise, the rm will be triggered right after the pod is launch successfully.
- the command line should be `/bin/sh -c "ps aux"`, rather than `/bin/sh -c ps aux`


gao-feng:
LGTM

PR #315, opened 2016-05-12
Add create/remove gRPC api for vm

feiskyer:
Part of #278


serverrpc/vm.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/golang/glog"
+	"github.com/hyperhq/hyperd/types"
+	"golang.org/x/net/context"
+)
+
+// VMCreate implements POST /vm/create
+func (s *ServerRPC) VMCreate(ctx context.Context, req *types.VMCreateRequest) (*types.VMCreateResponse, error) {
+	glog.V(3).Infof("VMCreate with request %s", req.String())
+
+	if req == nil || req.Cpu == 0 || req.Memory == 0 {
```
laijs: why "req == nil" is needed?

feiskyer: If req is indeed nil, then `req.Cpu == 0 || req.Memory == 0` will panic.

laijs: how to make it happen

feiskyer: The client may pass in a nil req.

Issue #283, opened 2016-04-15
Remove UserPod from runv


laijs:
And move runv.git/hypervisor/pod to hyper.git/pod/
UserPod is user interface and too high level.
It is better to send lower level and more verbose configuration to the APIs of the runv.
(currently verbose/additional configurations are send via additional arguments)

gnawux:
done in the refactor

PR #318, opened 2016-05-13
Add gPRC api for pod create/remove

feiskyer:
- Part of #278: Add gPRC api for pod create/remove
- Add integration test
- Ref #283: This PR also copy runv.UserPod to hyperd.


daemon/pod.go
```diff
@@ -42,11 +43,201 @@ type Pod struct {
 	sync.RWMutex
 }
 
-func NewPod(rawSpec []byte, id string, data interface{}) (*Pod, error) {
+// TODO: remove convertToRunvPodSpec after pod.UserPod is deleted from runv
+func convertToRunvPodSpec(podSpec *apitypes.UserPod) (*pod.UserPod, error) {
```
laijs: it is not RunvPodSpec. it is hyperhq's official pod spec: UserPod.
its code is just wrongly put at runv. it will be moved back to hyperd.
UserPod will not be removed, any attempt will be rejected by xu. it is already an inuse user interface.

feiskyer: This PR adds `UserPod` in hyperd, so the dedicated `pod.UserPod` could be called `RunvPodSpec` and `pod.UserPod` should be removed in another PR. 

There is a todo to remove this convertion: `// TODO: remove convertToRunvPodSpec after pod.UserPod is deleted from runv`

laijs: pod.UserPod is already user interface, it will not be removed(it will be moved back to hyperd)

convertToRunvPodSpec should be convertToUserPod for respecting the name.
I suggest to change apitypes.UserPod to apitypes.Pod

feiskyer: Not really. Apitypes.UserPod is the new interface in hyperd to replace runv's UserPod.

laijs: UserPod doesn't belong to runv, it is just wrongly put at runv. there is no "runv's UserPod".

pod.UserPod is user interface used by `hyperctl run -p` and persistent struct saved on database, it is almost pure hyperd api. but because the cleanup was not completed when the runv is separated from hyper, pod.UserPod is still inuse in runv. this history is also one of the reason why pod.UserPod should be removed from runv.

runv StartPod() needs an internal pod config for everything instead of pod.UserPod + additional config fragments. the internal pod config is much more changeable than any user api such as pod.UserPod. Apitypes.UserPod is also user api/interface, it will not be used on StartPod() even pod.UserPod is removed from runv.

laijs: apitypes.UserPod is pod config for network transport, it is user API/ABI. but it is not seen by user directly.
pod.UserPod is pod config which can be created manually by user for composing a pod.

I don't think apitypes.UserPod can replace pod.UserPod. and "UserPod" in apitypes.UserPod is confusing. apitypes.Pod is much better, like all other packagename.Pod.

feiskyer: No, `apitypes.UserPod` is for all hyperd APIs and clients. Users can see it directly:

- For gRPC api, users must specify `apitypes.UserPod` while creating Pods
- For original restful api, the `pod.json` is unmarshaled to `apitypes.UserPod`

After this PR, hyperd APIs will no longer depend on `UserPod` in runv. 

> I don't think apitypes.UserPod can replace pod.UserPod. 

Why can't?

types/types.proto
```diff
@@ -234,14 +234,138 @@ message VMRemoveResponse {
...
+  string path     = 1;
+  string filename = 2;
+  string perm     = 3;
+  string user     = 4;
+  string group    = 5;
+}
+
+message UserContainer {
+  string name                           = 1;
+  string image                          = 2;
```
laijs: in hypervisor/pod.go, there is `User          UserUser` field here.
we will have to add User as `UserUser user = 13` ?

types/types.proto
```diff
@@ -234,14 +234,138 @@ message VMRemoveResponse {
...
+  repeated UserContainer containers = 10;
+  repeated UserFile files           = 11;
+  repeated UserVolume volumes       = 12;
+  repeated UserInterface interfaces = 13;
+  repeated UserService services     = 14;
+}
+
+message PodCreateRequest {
+  UserPod podSpec = 1;
+  string podID    = 2;
```
laijs: which id will be  used? this podID or podSpec.Id?

daemon/pod.go
gao-feng: apitypes.UserPod and runv.UserPod(shoud be hyperd.UserPod) are both user faced API structure. One for grpc/protobuf, One for json. they will be converted to runv.PodInfo(doesn't exist yet), they are two types of structure services for two types of API. 

Since the runv.UserPod represents the pod spec right now(in json), If user uses grpc api and pod spec to create pod,  the extra conversion from json to protobuf is needed.

feiskyer: > If user uses grpc api and pod spec to create pod, the extra conversion from json to protobuf is needed.

No, there is no json to protobuf conversion in gRPC api. Users could pass apitypes.UserPod struct directly in create API.

gao-feng: I mean If hyprectl use grpc someday, the command `hyperctl run -p pod.spc`  will need to do conversion

feiskyer: Yes, if hyperctl switches to gRPC, then the conversion should be done on hyperctl client.

hack/lib/test.sh
```diff
@@ -37,7 +37,7 @@ hyper::test::exec() {
   echo "Pod exec and exit code test"
   id=$(sudo hyperctl run -t -d busybox /bin/sh | sed -ne "s/POD id is \(pod-[0-9A-Za-z]\{1,\}\)/\1/p")
   echo "test pod ID is $id"
-  res=$(sudo hyperctl exec $id sh -c "exit 37" > /dev/null 2>&1 ; echo $?)
+  res=$(sudo hyperctl exec $id sh -c "exit 37" ; echo $?)
```
laijs: it needs to be split into 2 lines.
otherwise $res may contains other garbarge

PR #324, opened 2016-05-19
Add gRPC api for container logs

YaoZengzeng:
Container logs gRPC api implemented by reading logs line by line.Each time after reading one line,we put it to the stream.
There is a little problem: we put every line of log in the type of string, but sometimes the logs may have a prefix(8 bytes maybe) for multiplexing, print it directly may not be perfect.So we have to process the prefix by the client.


serverrpc/logs.go
```diff
@@ -0,0 +1,62 @@
...
+		s, n, err := timetypes.ParseTimestamps(req.Since, 0)
+		if err != nil {
+			return err
+		}
+		since = time.Unix(s, n)
+	}
+
+	buffer := bytes.NewBuffer([]byte{})
+
+	var stop <-chan bool
```
laijs: `stop` is nil channel here, you should create one via:
`stop := make(chan bool, 1)`

serverrpc/logs.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	err := s.daemon.GetContainerLogs(req.Container, logsConfig)
+	if err != nil {
+		glog.Errorf("ContainerLogs error: %v", err)
+		return err
+	}
+
+	for {
+		s, err := buffer.ReadString(10)
```
laijs: use `ReadString('\n')`

types/types.proto
```diff
@@ -356,6 +356,21 @@ message PodRemoveResponse {
   string cause  = 2;
 }
 
+message ContainerLogsRequest {
+  string  container =1;
+  bool    follow    =2;
+  bool    timestamps=3;
```
laijs: comparing to existing definition, `bool    timestamps = 3;` seems better.
there is one blank at least before `=` and one blank exactly after `=`.

types/types.proto
```diff
@@ -356,6 +356,21 @@ message PodRemoveResponse {
...
+  string  tail      =4;
+  string  since     =5;  
+  bool    stdout    =6;
+  bool    stderr    =7;
+
+}
+
+message ContainerLogsResponse {
+  string  log      =1;
```
feiskyer: Should use bytes here to avoid conversions.

YaoZengzeng: Yes,I've tried it with "repeated byte",but there seems no tyep as "byte".And "bytes" is OK

feiskyer: No need for repeated, `bytes` is mapping to `[]byte` in Go,

serverrpc/logs.go
feiskyer: What will happen if there is no `\n` in logs? Will it block for ever?

feiskyer:
@YaoZengzeng To pass travis, need run `hack/update-generated-proto.sh`.

serverrpc/logs.go
YaoZengzeng: No, it will return the whole bytes in buffer,and "err = io.EOF".And I have to process it and return the left bytes without '\n'.But maybe we can just read the logs at once,like `rpc ContainerLogs(ContainerLogsRequest) returns (ContainerLogsResponse) {}`?

serverrpc/logs.go
```diff
@@ -0,0 +1,74 @@
...
+			if logsConfig.Follow == false {
+				eof = true
+			}
+		} else if err != nil {
+			glog.Errorf("Read log stream error: %v", err)
+			return err
+		}
+
+		if err := stream.Send(&types.ContainerLogsResponse{Log: s}); err != nil {
+			stop <- true
```
YaoZengzeng: The channel "stop" seems useless, it doesn't be used anywhere in `daemon.GetContainerLogs`?
And I want to use it to stop the `daemon.GetContainerLogs` if I dont want to get logs anymore.
May I use the "stop" in `daemon.GetContainerLogs` to notify the method to stop ?

types/types.proto
```diff
@@ -356,6 +356,21 @@ message PodRemoveResponse {
...
 
+message ContainerLogsRequest {
+  string  container   = 1;
+  bool    follow      = 2;
+  bool    timestamps  = 3;
+  string  tail        = 4;
+  string  since       = 5;  
+  bool    stdout      = 6;
+  bool    stderr      = 7;
+
```
feiskyer: Remove new lines here

serverrpc/logs.go
feiskyer: Yes, that is what we should do.

feiskyer:
LGTM. Thanks.

PR #329, opened 2016-05-23
add err condition avoid panic

Crazykev:
Get panic when I run command `hyperctl images` with non-privileged user. 
Fix it.


gnawux:
LGTM, thank you @Crazykev

PR #332, opened 2016-05-25
update protnbuf dependency package version to version 2

gnawux:
try to fix the build issue caused by protobuf update


feiskyer:
Please run hack/update-generated-proto.sh, do not edit types.pb.go manually.

PR #339, opened 2016-05-26
Add gRPC API for exitcode

Crazykev:
Add gRPC API for exitcode
part of #278 


integration/client.go
```diff
@@ -41,6 +41,20 @@ func (c *HyperClient) GetPodInfo(podID string) (*types.PodInfo, error) {
 	return pod.PodInfo, nil
 }
 
+// ExitCode gets exitcode by container and tag
+func (c *HyperClient) ExitCode(container, tag string) (int32, error) {
+	request := types.ExitCodeRequest{
+		Container: container,
+		Tag:       tag,
```
laijs: tag is going to be removed

Crazykev: @laijs So how can we decide which tty without a client tag?

laijs: ExitCode is for process(container's init process or exec process). the tag will be replaced by process Id.

Crazykev: got it. I'll update this patch until related patch is merged.

feiskyer: @Crazykev we can make a workaround for now: Rename `ExitCode` to `Wait`, and rename `tag` to id.

Crazykev: yep, updated and rebased @feiskyer

types/types.proto
```diff
@@ -366,10 +366,20 @@ message ContainerLogsRequest {
   bool    stderr      = 7;
 }
 
+
```
laijs: why there is an additional empty line here?

types/types.proto
```diff
@@ -366,10 +366,20 @@ message ContainerLogsRequest {
...
 message ContainerLogsResponse {
   bytes log = 1;
 }
 
+message WaitRequest{
+    string container = 1;
+    string processId = 2;
+}
```
laijs: it seems better if there is a `bool noHang = 3`.
it can be added later if it is needed.

Crazykev:
@laijs updated and rebased

PR #340, opened 2016-05-26
do these checks firstly when we start hyperd

xlgao-zju:
do these checks firstly when we start hyperd. If these checks fail, we don`t need to do other things.


xlgao-zju:
ping @carmark

carmark:
LGTM

PR #342, opened 2016-05-27
add grpc info api

xlgao-zju:
A part of #278 


types/types.proto
```diff
@@ -370,6 +370,24 @@ message ContainerLogsResponse {
   bytes log = 1;
 }
 
+message InfoRequest {
+}
+
+message InfoResponse {
+  string  ID                    = 1;
+  string  Containers            = 2;
```
laijs: its type is int32

laijs:
change the return type of CmdSystemInfo() to be types.InfoResponse for new gRPC api.

and convert it into engine.Env for old api.
// old api: server/router/system/system_routes.go:18
```go
func (s *systemRouter) getInfo(ctx context.Context, w http.ResponseWriter, r *http.Request, vars map[string]string) error {
        env, err := s.backend.CmdSystemInfo()
        if err != nil {
                return err
        }

        return env.WriteJSON(w, http.StatusOK)
}
```

types/types.proto
```diff
@@ -370,6 +370,24 @@ message ContainerLogsResponse {
...
+  string  Images                = 3;
```
laijs: int32

types/types.proto
```diff
@@ -370,6 +370,24 @@ message ContainerLogsResponse {
...
+  string  Driver                = 4;
+  string  DriverStatus          = 5;
```
laijs: define a grpc struct

types/types.proto
```diff
@@ -370,6 +370,24 @@ message ContainerLogsResponse {
...
+  string  DockerRootDir         = 6;
+  string  IndexServerAddress    = 7;
+  string  ExecutionDriver       = 8;
+  string  MemTotal              = 9;
+  string  Pods                  = 10;
```
laijs: int32

types/types.proto
```diff
@@ -370,6 +370,24 @@ message ContainerLogsResponse {
...
+message InfoResponse {
+  string  ID                    = 1;
+  string  Containers            = 2;
+  string  Images                = 3;
+  string  Driver                = 4;
+  string  DriverStatus          = 5;
+  string  DockerRootDir         = 6;
+  string  IndexServerAddress    = 7;
+  string  ExecutionDriver       = 8;
+  string  MemTotal              = 9;
```
laijs: uint64 ?

xlgao-zju:
OK, I will do it later.

types/types.proto
xlgao-zju: this one is int64

xlgao-zju:
@laijs updated

feiskyer:
LGTM

PR #344, opened 2016-05-28
Add gRPC api for attach

YaoZengzeng:
ref #278 attach api


types/types.proto
```diff
@@ -394,6 +394,13 @@ message InfoResponse {
   string  name                    = 12;
 }
 
+message AttachMessage {
+  string key  = 1;
```
feiskyer: Remove key as we don't really need it.

types/types.proto
```diff
@@ -394,6 +394,13 @@ message InfoResponse {
...
+  string id   = 2;
```
feiskyer: rename this field to containerID

feiskyer:
lgtm

PR #349, opened 2016-06-01
add grpc api for exec

Crazykev:
part of #278 


integration/client.go
```diff
@@ -214,3 +215,72 @@ func (c *HyperClient) RemovePod(podID string) error {
 
 	return nil
 }
+
+// ContainerExec exec a command in a container with input stream in and output stream out
+func (c *HyperClient) ContainerExec(container, tag string, command []byte, tty bool, stdin io.ReadCloser, stdout, stderr io.Writer) error {
```
feiskyer: Just `command string` here, avoid unnecessary type conversions.

integration/client.go
```diff
@@ -214,3 +215,72 @@ func (c *HyperClient) RemovePod(podID string) error {
...
+	request := types.ContainerExecRequest{
+		Type:    "container",
+		Value:   container,
+		Command: string(command),
```
feiskyer: Command should be json marshaled:

```go
cmd, err := json.Marshal(command)
    if err != nil {
        return err
    }
```

And then set `Command: string(cmd)`

Crazykev:
@feiskyer move `command` marshal to server. updated.

feiskyer:
LGTM

PR #355, opened 2016-06-05
Refactoring GetByName in PodList  Changed return types of GetByName to keep consistency with other Get* functions

daehyeok:


laijs:
LGTM

PR #365, opened 2016-06-09
Support for stopping container

YaoZengzeng:
Just directly use kill SIGKILL to stop the container for temporary.
And modify "stop command of hyperctl" like "kill" which can stop pod or container.


daemon/stop.go
```diff
@@ -89,3 +90,21 @@ func (daemon *Daemon) StopPodWithinLock(pod *Pod) (int, string, error) {
...
+	containerId := p.status.Containers[idx].Id
+	glog.V(1).Infof("found container %s to stop", containerId)
+
+	err := p.vm.KillContainer(containerId, syscall.SIGTERM)
+	if err == nil {
+		return nil
+	}
+	err = p.vm.KillContainer(containerId, syscall.SIGKILL)
+	return err
+	//TODO: update container status of the pod
```
feiskyer: @laijs  Any updates on pod/container state machine?

types/types.proto
```diff
@@ -541,6 +549,8 @@ service PublicAPI {
     // TODO: ContainerCommit commits the changes of the specified container
     // TODO: ContainerSignal sends a singla to specified container
     // TODO: ContainerLabels updates labels of the specified container
+    // ContainerStop implements POST /container/stop
```
feiskyer: ContainerStop stops the specified container

feiskyer: `/container/stop` restful api and client are also required.

daemon/stop.go
YaoZengzeng: ref to runv pr #251

daemon/stop.go
```diff
@@ -89,3 +90,21 @@ func (daemon *Daemon) StopPodWithinLock(pod *Pod) (int, string, error) {
...
+		return fmt.Errorf("can not found container %s", container)
+	}
+
+	containerId := p.status.Containers[idx].Id
+	glog.V(1).Infof("found container %s to stop", containerId)
+
+	err := p.vm.KillContainer(containerId, syscall.SIGTERM)
+	if err == nil {
+		return nil
+	}
```
laijs: p.vm.KillContainer(containerId, syscall.SIGTERM) hardly fail. but the process may continue to run.
please wait/check the container's status(we haven't had such api) in several seconds and then kill SIGKILL if needed.
or directly kill SIGKILL.

daemon/stop.go
```diff
@@ -89,3 +90,17 @@ func (daemon *Daemon) StopPodWithinLock(pod *Pod) (int, string, error) {
...
+	if !ok {
+		return fmt.Errorf("can not found container %s", container)
+	}
+
+	containerId := p.status.Containers[idx].Id
+	glog.V(1).Infof("found container %s to stop", containerId)
+
+	err := p.vm.KillContainer(containerId, syscall.SIGKILL)
+	return err
+	//TODO: update container status of the pod
```
feiskyer: Could you update the status of container and keep pod running even container is stopped?

YaoZengzeng: Yes,I'll do it in next pr.
Just call  "SetOneContainerStatus" in runv is fine.

daemon/stop.go
```diff
@@ -89,3 +90,17 @@ func (daemon *Daemon) StopPodWithinLock(pod *Pod) (int, string, error) {
...
+func (daemon *Daemon) StopContainer(container string) error {
+	p, idx, ok := daemon.PodList.GetByContainerIdOrName(container)
+	if !ok {
+		return fmt.Errorf("can not found container %s", container)
+	}
+
+	containerId := p.status.Containers[idx].Id
+	glog.V(1).Infof("found container %s to stop", containerId)
+
+	err := p.vm.KillContainer(containerId, syscall.SIGKILL)
```
gao-feng: you should check if p.vm is nil (pod is not started yet)

client/stop.go
```diff
@@ -12,10 +12,11 @@ import (
 func (cli *HyperClient) HyperCmdStop(args ...string) error {
 
 	var opts struct {
+		Pod  bool `short:"p" long:"pod" default:"false" default-mask:"-" description:"stop a pod"`
```
gao-feng: Should stop pod by default.

daemon/stop.go
```diff
@@ -89,3 +90,21 @@ func (daemon *Daemon) StopPodWithinLock(pod *Pod) (int, string, error) {
 
 	return vmResponse.Code, vmResponse.Cause, nil
 }
+
+func (daemon *Daemon) StopContainer(container string) error {
+	pod, idx, ok := daemon.PodList.GetByContainerIdOrName(container)
+	if !ok {
+		return fmt.Errorf("can not found container %s", container)
+	}
+
```
YaoZengzeng: Transition Lock may need there.But I'll do it in the next pr which update  container status of the pod.

feiskyer:
lgtm

PR #375, opened 2016-06-23
Implement rpc for SetPodLabels and PodStats

resouer:
gRpc implementation for:

1. SetPodLabels
2. PodStats

Use `io.Pipe` to replace `buffer` in ImagePull, since there's data race and caused nil pointer in buffer Truncate

### DIFF
--- a/integration/client.go
+++ b/integration/client.go
@@ -534,6 +534,20 @@ func (c *HyperClient) ListService(podID string) ([]*types.UserService, error) {
 	return resp.Services, nil
 }
 
+// GetPodStats get stats of Pod by podID
+func (c *HyperClient) GetPodStats(podID string) (*types.PodStats, error) {
+	statsResponse, err := c.client.PodStats(
+		c.ctx,
+		&types.PodStatsRequest{PodID: podID},
+	)
+
+	if err != nil {
+		return nil, err
+	}
+
+	return statsResponse.PodStats, nil
+}
+
 // AddService adds user service by podID and service content
 func (c *HyperClient) AddService(podID string, services []*types.UserService) error {
 	_, err := c.client.ServiceAdd(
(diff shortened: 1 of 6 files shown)


types/types.proto
```diff
@@ -610,6 +744,12 @@ service PublicAPI {
     // VMRemove deletes a HyperVM by vmID
     rpc VMRemove(VMRemoveRequest) returns (VMRemoveResponse) {}
 
+    // SetPodLabels sets labels of given pod
+    rpc SetPodLabels(PodLabelsRequest) returns (PodLabelsResponse) {}
+
+    // GetPodStats implements  GET    /pod/stats
```
feiskyer: Follow docs above.

types/types.proto
```diff
@@ -572,6 +691,23 @@ message PodUnpauseRequest {
...
 message PodUnpauseResponse {}
 
+message PodLabelsRequest{
+  string podID               = 1;
+  bool override              = 2;
+  map<string, string> labels = 3;
+}
+
+message PodLabelsResponse{
+}
```
feiskyer: It's better to keep `{}` in one line same as other parts.

integration/client.go
```diff
@@ -336,6 +336,7 @@ func (c *HyperClient) ContainerExec(container, tag string, command []string, tty
 // StartPod starts a pod by podID
 func (c *HyperClient) StartPod(podID, vmID, tag string) error {
 	stream, err := c.client.PodStart(context.Background())
+
```
feiskyer: New line is not needed here.

integration/hyper_test.go
```diff
@@ -404,3 +432,30 @@ func (s *TestSuite) TestPauseAndUnpausePod(c *C) {
 	err = s.client.RemovePod(pod)
 	c.Assert(err, IsNil)
 }
+
+func (s *TestSuite) TestGetPodStats(c *C) {
```
feiskyer: `GetPodStats` only works for `libvirt`, but the tests all cover `qemu` hypervisor. Could you make this test only runs for `libvirt`? 

The hypervisor type could be got by calling `Info` api:

```
hyperctl info | grep Execu
Execution Driver: libvirt
```

serverrpc/pod.go
```diff
@@ -163,3 +164,190 @@ func (s *ServerRPC) PodUnpause(ctx context.Context, req *types.PodUnpauseRequest
...
+		glog.Errorf("PodLabels error: %v", err)
+		return nil, err
+	}
+
+	return &types.PodLabelsResponse{}, nil
+}
+
+// PodStats get stats (runvtypes.PodStats) of Pod
+func (s *ServerRPC) PodStats(c context.Context, req *types.PodStatsRequest) (*types.PodStatsResponse, error) {
+	glog.V(3).Infof("GEt pod stats with request %v", req.String())
```
feiskyer: s/GEt/get/

types/types.proto
```diff
@@ -610,6 +743,12 @@ service PublicAPI {
     // VMRemove deletes a HyperVM by vmID
     rpc VMRemove(VMRemoveRequest) returns (VMRemoveResponse) {}
 
+    // SetPodLabels sets labels of given pod
+    rpc SetPodLabels(PodLabelsRequest) returns (PodLabelsResponse) {}
+
+    // GetPodStats get pod stats of a given pod
```
feiskyer: s/GetPodStats/PodStats/, for keeping doc consistent with real interface name.

feiskyer:
LGTM. Thanks.

PR #378, opened 2016-06-27
optimize get pod by name

gnawux:
add a name to pod index in podlist

### DIFF
--- a/daemon/podlist.go
+++ b/daemon/podlist.go
@@ -10,6 +10,7 @@ import (
 
 type PodList struct {
 	pods       map[string]*Pod
+	names      map[string]*Pod
 	containers map[string]string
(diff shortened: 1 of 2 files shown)


daemon/podlist.go
```diff
@@ -63,18 +67,11 @@ func (pl *PodList) GetByName(name string) (*Pod, bool) {
...
-	pod := pl.findUnsafe(func(p *Pod) bool {
-		if p.status.Name == name {
-			return true
-		}
-		return false
-	})
-
-	if pod != nil {
-		return pod, true
+	if pl.names == nil {
```
laijs: pl.names is not possible to be nil

gnawux:
this does not change any result, close it.

Issue #384, opened 2016-06-29
Extend info api for pod and container


feiskyer:
Following fields are required:

- PodInfo and PodList
  - CreatedAt
  - PodID
  - PodName
  - Labels
- ContainerInfo
  - CreatedAt

PR #386, opened 2016-06-29
Extend pod and container info

feiskyer:
Implements #384

### DIFF
--- a/daemon/pod.go
+++ b/daemon/pod.go
@@ -31,13 +31,23 @@ import (
 	"github.com/hyperhq/runv/hypervisor/types"
 )
 
+type Container struct {
+	ApiContainer *apitypes.Container
+	mountID      string
+	fstype       string
+	rootfs       string
+	initialize   bool
+}
+
 type Pod struct {
-	Id           string
-	status       *hypervisor.PodStatus
-	spec         *pod.UserPod
-	vm           *hypervisor.Vm
-	ctnStartInfo []*hypervisor.ContainerInfo
-	volumes      map[string]*hypervisor.VolumeInfo
+	Id        string
+	CreatedAt int64
+	PodStatus *hypervisor.PodStatus
+	Spec      *pod.UserPod
+	VM        *hypervisor.Vm
+
+	containers []*Container
+	volumes    map[string]*hypervisor.VolumeInfo
 
 	transiting chan bool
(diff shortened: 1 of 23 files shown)


daemon/pod.go
```diff
@@ -31,12 +31,21 @@ import (
...
 	Id           string
-	status       *hypervisor.PodStatus
-	spec         *pod.UserPod
-	vm           *hypervisor.Vm
-	ctnStartInfo []*hypervisor.ContainerInfo
+	CreatedAt    int64
+	PodStatus    *hypervisor.PodStatus
+	Spec         *pod.UserPod
+	VM           *hypervisor.Vm
+	ctnStartInfo []*Container
```
laijs: it should be Containers rather than ctnStartInfo

daemon/pod.go
```diff
@@ -285,14 +286,15 @@ func NewPod(podSpec *apitypes.UserPod, id string, data interface{}) (*Pod, error
 
 	p := &Pod{
 		Id:         id,
+		CreatedAt:  time.Now().Unix(),
```
laijs: it might be a different value when it is call from `dameon.Restore() -> daemon.createPodInternal() -> .... `

daemon/storage.go
```diff
@@ -158,11 +158,11 @@ func (dms *DevMapperStorage) CleanUp() error {
...
 }
 
-func (dms *DevMapperStorage) PrepareContainer(ci *hypervisor.ContainerInfo, sharedDir string) error {
-	if err := dm.CreateNewDevice(ci.MountId, dms.DevPrefix, dms.RootPath()); err != nil {
+func (dms *DevMapperStorage) PrepareContainer(ci *Container, sharedDir string) error {
+	if err := dm.CreateNewDevice(ci.mountID, dms.DevPrefix, dms.RootPath()); err != nil {
 		return err
 	}
-	devFullName, err := dm.MountContainerToSharedDir(ci.MountId, sharedDir, dms.DevPrefix)
+	devFullName, err := dm.MountContainerToSharedDir(ci.ApiContainer.ContainerID, sharedDir, dms.DevPrefix)
```
gao-feng: why change mountid to containerid?

feiskyer: My mistake. Fixed.

PR #391, opened 2016-07-09
Remove "vm-XXX" directory when vm exits

YaoZengzeng:


laijs:
it is better to move the sharedir out from vmdir.
ctx.Close() removes the vmdir
pod.Cleanup() removes the sharedir

This patch works now, I will merge it (except the VMs has no pod).
But could you create PRs to implement above comments?

PS:
In the future, all the sharedir management is moved to runv. After that
sharedir can be in the vmdir.

YaoZengzeng:
OK， I‘ll do it.

Issue #319, opened 2016-05-15
Stop hyper by pod-name


haosdent:
Now hyper stop command is 

```
$ hyper stop --help
Usage:
  hyper stop POD_ID
```

Could it support to use `pod-name` as the parameter?

gnawux:
Will support this soon. Thank you @haosdent

gnawux:
supported in latest version

PR #398, opened 2016-07-27
Enable using pod-name in hyperctl commands

hzmangel:
This PR enables using pod-name in hyperctl commands, which contains those changes:
- A new function in `PodList` struct, which chooses correct scope (Id or Name) to query `Pod` object based on prefix.
- Replace `PodList.Get` in commands to new function.

Please help to review, thanks.


hzmangel:
ref #319

Jimmy-Xu:

Jimmy-Xu:

gnawux:
I think this has already included in the latest master

PR #400, opened 2016-08-04
add container rename grpc api

YaoZengzeng:


laijs:
could you add the client code or the test code in a new commit in this pr?

Issue #381, opened 2016-06-28
the options of the APP are eaten if the options are also the options for `hyperctl run/exec`


laijs:
`hyperctl run busybox uname -a`

the `-a` will be eaten.

`hyperctl run busybox uname -t` should be error, but `-t` is eaten and avoid the error.

gnawux:
with `gflag.PassAfterNonOption`, the flags after first non-option arg will be pass as arguments. It has been included in commit 7481c54f0ebed4d3019adf5c3679d4512803e27d

laijs:
@gnawux gflag.PassAfterNonOption is not added to `hyperctl exec` yet

gnawux:
add it to exec in #561

gnawux:
already fixed in the latest release

PR #403, opened 2016-08-10
[hyperctl] Pass all args after image name in `run'

hzmangel:
The default go-flags parses all args with dash prefix, includes args for running command. This fix uses an extra function to get all args after image name to solve this problem.

This fix can solve problem mentioned in #381. However, the `go-flags` package still block some flags, such as `--help`. Will find solution for this later.


laijs:
the go-flags parses and eats some flags after the imagename, this patch just recover the eaten flags. we hope the solution make go-flags stop parsing at the imagename, or change go-flags to other flags-parsing package.

We are looking for a full solution, thanks.

PR #410, opened 2016-08-18
change response code to -2, when we can not find the Pod

xlgao-zju:
As @feiskyer said in this [pr](https://github.com/kubernetes/frakti/pull/11), I think we should change this response code.

### DIFF
--- a/daemon/rm.go
+++ b/daemon/rm.go
@@ -11,21 +11,27 @@ import (
 	"github.com/hyperhq/runv/hypervisor/types"
 )
 
+const (
+	E_NOT_FOUND       = -2
+	E_UNDER_OPERATION = -1
+	E_OK              = 0
+)
+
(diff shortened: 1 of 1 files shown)


daemon/rm.go
```diff
@@ -20,7 +20,7 @@ func (daemon *Daemon) CleanPod(podId string) (int, string, error) {
 
 	pod, ok := daemon.PodList.Get(podId)
 	if !ok {
-		return -1, "", fmt.Errorf("Can not find that Pod(%s)", podId)
+		return -2, "", fmt.Errorf("Can not find that Pod(%s)", podId)
```
feiskyer: Could you define `-1` and `-2` as const?

xlgao-zju:
Code updated.

feiskyer:
lgtm

PR #413, opened 2016-08-26
Randomize service container name

resouer:
Rename service container to random string to avoid conflicts after restarted hyperd.

cc @feiskyer

### DIFF
--- a/daemon/list.go
+++ b/daemon/list.go
@@ -2,6 +2,7 @@ package daemon
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/hyperhq/runv/hypervisor"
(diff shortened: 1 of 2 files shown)


gnawux:
Agree. I have the same idea when I refactor the hyperd,  all determined name might be collision.

feiskyer:
@resouer This will introduce problems on https://github.com/hyperhq/hyperd/blob/master/daemon/list.go#L107, maybe change `filterPodContainers` logic with suffix instead of name?

resouer:
Do not merge! list will broke, I have noticed that.

feiskyer:
lgtm

PR #414, opened 2016-08-27
[hyperctl / API] Add -q params to `hyperctl list` command

hzmangel:
While running `docker ps` command, there is a params called `-q`, that only list container ids, without header and other information. The pure id list can be added to some other commands easily, such as `docker stop $(docker ps -aq)`.

This PR added `-q` to `hyperctl list` command to simulate the function. With this params, hyperctl list only shows id for VM / Pod / Container. Please help to review, thanks.

### DIFF
--- a/client/api/interface.go
+++ b/client/api/interface.go
@@ -18,7 +18,7 @@ type APIInterface interface {
 
 	WinResize(id, tag string, height, width int) error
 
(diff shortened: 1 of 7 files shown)


Jimmy-Xu:

Jimmy-Xu:

PR #423, opened 2016-09-15
fix output bug of grpc attach

YaoZengzeng:
fix the bug that sometimes can't get the output of grpc attach and make the code more robust.

Signed-off-by: Yao Zengzeng <yaozengzeng@zju.edu.cn>

### DIFF
--- a/serverrpc/attach.go
+++ b/serverrpc/attach.go
@@ -22,23 +22,42 @@ func (s *ServerRPC) Attach(stream types.PublicAPI_AttachServer) error {
 	go func() {
 		for {
 			cmd, err := stream.Recv()
-			if err != nil {
+			if err == io.EOF {
 				return
 			}
-			if _, err := iw.Write(cmd.Data); err != nil {
+			if err != nil {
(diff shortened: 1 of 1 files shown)


serverrpc/attach.go
```diff
@@ -22,23 +22,41 @@ func (s *ServerRPC) Attach(stream types.PublicAPI_AttachServer) error {
...
-			if _, err := iw.Write(cmd.Data); err != nil {
+
+			n, err := iw.Write(cmd.Data)
+			if err != nil {
+				glog.Errorf("Write pipe error: %v", err)
 				return
 			}
+			if n != len(cmd.Data) {
+				glog.Errorf("Write data length is not enough, write: %d, success: %d", len(cmd.Data), n)
+			}
```
gnawux: Here you spot an error, but won't interrupt the stream. Is this what you want exactly?

gnawux:
LGTM

This is a minor fix and blocks k8s 1.3 integration. Merge it.

PR #429, opened 2016-09-26
add --volume support for hyperctl

### LINKED ISSUE #341: add volume support in `hyperctl run` command line

crook:
use bind mount to close #341 

Now support: hyperctl run -v /host-src:/container-dest -t <images>

Test log:
http://pastebin.com/Wr1mZwEa

### DIFF
--- a/client/run.go
+++ b/client/run.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"path/filepath"
 	"regexp"
 	"strconv"
 	"strings"
@@ -40,6 +41,7 @@ func (cli *HyperClient) HyperCmdRun(args ...string) (err error) {
 		Remove        bool     `long:"rm" default:"false" default-mask:"-" description:"Automatically remove the pod when it exits"`
 		Portmap       []string `long:"publish" value-name:"[]" default-mask:"-" description:"Publish a container's port to the host, format: --publish [tcp/udp:]hostPort:containerPort"`
 		Labels        []string `long:"label" value-name:"[]" default-mask:"-" description:"Add labels for Pod, format: --label key=value"`
+		Volumes       []string `short:"v" long:"volume" value-name:"[]" default-mask:"-" description:"Mount host file/directory as a data file/volume, format: -v|--volume=[[hostDir:]containerDir[:options]]"`
 	}
 
 	var (
@@ -71,7 +73,7 @@ func (cli *HyperClient) HyperCmdRun(args ...string) (err error) {
 		}
 		attach = !opts.Detach
 		podJson, err = cli.JsonFromCmdline(args, opts.Env, opts.Portmap, opts.LogDriver, opts.LogOpts,
-			opts.Name, opts.Workdir, opts.RestartPolicy, opts.Cpu, opts.Memory, opts.Tty, opts.Labels, opts.EntryPoint)
+			opts.Name, opts.Workdir, opts.RestartPolicy, opts.Cpu, opts.Memory, opts.Tty, opts.Labels, opts.EntryPoint, opts.Volumes)
 	}
 
 	if err != nil {
@@ -196,16 +198,18 @@ func (cli *HyperClient) JsonFromFile(filename string, yaml, k8s bool) (string, e
 }
 
 func (cli *HyperClient) JsonFromCmdline(cmdArgs, cmdEnvs, cmdPortmaps []string, cmdLogDriver string, cmdLogOpts []string,
-	cmdName, cmdWorkdir, cmdRestartPolicy string, cpu, memory int, tty bool, cmdLabels []string, entrypoint string) (string, error) {
+	cmdName, cmdWorkdir, cmdRestartPolicy string, cpu, memory int, tty bool, cmdLabels []string, entrypoint string, cmdVols []string) (string, error) {
 
 	var (
-		name    = cmdName
-		image   = cmdArgs[0]
-		command = []string{}
-		env     = []pod.UserEnvironmentVar{}
-		ports   = []pod.UserContainerPort{}
-		logOpts = make(map[string]string)
-		labels  = make(map[string]string)
+		name       = cmdName
+		image      = cmdArgs[0]
+		command    = []string{}
+		env        = []pod.UserEnvironmentVar{}
+		ports      = []pod.UserContainerPort{}
+		logOpts    = make(map[string]string)
+		labels     = make(map[string]string)
+		volumesRef = []pod.UserVolumeReference{}
+		volumes    = []pod.UserVolume{}
 	)
 	if len(cmdArgs) > 1 {
 		command = cmdArgs[1:]
@@ -254,6 +258,15 @@ func (cli *HyperClient) JsonFromCmdline(cmdArgs, cmdEnvs, cmdPortmaps []string, 
 		}
 	}
 
+	for _, v := range cmdVols {
+		vol, volRef, err := parseVolume(v)
+		if err != nil {
+			return "", err
+		}
+		volumes = append(volumes, *vol)
+		volumesRef = append(volumesRef, *volRef)
+	}
+
 	entrypoints := make([]string, 0, 1)
 	if len(entrypoint) > 0 {
 		entrypoints = append(entrypoints, entrypoint)
@@ -267,7 +280,7 @@ func (cli *HyperClient) JsonFromCmdline(cmdArgs, cmdEnvs, cmdPortmaps []string, 
 		Entrypoint:    entrypoints,
 		Ports:         ports,
 		Envs:          env,
-		Volumes:       []pod.UserVolumeReference{},
+		Volumes:       volumesRef,
 		Files:         []pod.UserFileReference{},
 		RestartPolicy: cmdRestartPolicy,
 	}}
@@ -278,7 +291,7 @@ func (cli *HyperClient) JsonFromCmdline(cmdArgs, cmdEnvs, cmdPortmaps []string, 
 		Labels:     labels,
 		Resource:   pod.UserResource{Vcpu: cpu, Memory: memory},
 		Files:      []pod.UserFile{},
-		Volumes:    []pod.UserVolume{},
+		Volumes:    volumes,
 		LogConfig: pod.PodLogConfig{
 			Type:   cmdLogDriver,
 			Config: logOpts,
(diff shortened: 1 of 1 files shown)


client/run.go
```diff
@@ -290,6 +315,50 @@ func (cli *HyperClient) JsonFromCmdline(cmdArgs, cmdEnvs, cmdPortmaps []string,
...
+		// cmd: -v host-src:container-dest
+		srcName = fields[0]
+		destPath = fields[1]
+		readOnly = "rw"
+	} else if len(fields) == 1 {
+		// -v container-dest
+		destPath = fields[0]
+		readOnly = "rw"
+		_, filename = path.Split(destPath)
+		srcName = path.Join("/var/tmp/hyper", filename+utils.RandStr(5, "number"))
```
laijs: please set the srcName to be "" in this case

client/run.go
```diff
@@ -254,6 +258,27 @@ func (cli *HyperClient) JsonFromCmdline(cmdArgs, cmdEnvs, cmdPortmaps []string,
...
+	for _, v := range cmdVols {
+		volRef, err := parseVolumes(v)
+		if err != nil {
+			return "", err
+		}
+		// Bind mount's driver is 'vfs' so far
+		var vol = pod.UserVolume{}
+		vol.Name = volRef.Volume
+		vol.Source = volRef.Volume
+		vol.Driver = "vfs"
```
laijs: when vol.Source is "", vol.Driver should be "", the hyperd will use its default volume driver

client/run.go
```diff
@@ -290,6 +315,50 @@ func (cli *HyperClient) JsonFromCmdline(cmdArgs, cmdEnvs, cmdPortmaps []string,
 	return string(jsonString), nil
 }
 
+func parseVolumes(volStr string) (*pod.UserVolumeReference, error) {
```
laijs: it seams better that it also returns the pod.UserVolume.

client/run.go
```diff
@@ -290,6 +315,50 @@ func (cli *HyperClient) JsonFromCmdline(cmdArgs, cmdEnvs, cmdPortmaps []string,
...
+		readOnly string
+		filename string
+	)
+
+	fields := strings.Split(volStr, ":")
+	if len(fields) == 3 {
+		// cmd: -v host-src:container-dest:rw
+		srcName = fields[0]
+		destPath = fields[1]
+		readOnly = fields[2]
```
laijs: please check the readOnly here.

crook:
@laijs @gnawux please help check the new version. Thanks

client/run.go
```diff
@@ -290,6 +303,66 @@ func (cli *HyperClient) JsonFromCmdline(cmdArgs, cmdEnvs, cmdPortmaps []string,
...
+		// -v container-dest
+		srcName = ""
+		destPath = fields[0]
+		readOnly = "rw"
+	} else {
+		return nil, nil, fmt.Errorf("flag format should be like : --volume=[host-src:]container-dest[:rw|ro]")
+	}
+
+	vol.Source = srcName
+	_, vol.Name = path.Split(vol.Source)
```
laijs: it might need to be a name with order as suffix or a random name

client/run.go
```diff
@@ -290,6 +303,66 @@ func (cli *HyperClient) JsonFromCmdline(cmdArgs, cmdEnvs, cmdPortmaps []string,
...
+			return nil, nil, fmt.Errorf("flag only support(ro or rw): --volume")
+		}
+	} else if len(fields) == 2 {
+		// cmd: -v host-src:container-dest
+		srcName = fields[0]
+		destPath = fields[1]
+		readOnly = "rw"
+	} else if len(fields) == 1 {
+		// -v container-dest
+		srcName = ""
```
laijs: in golang, "" is the initial/default value of a string. please remove this line.

client/run.go
```diff
@@ -290,6 +303,66 @@ func (cli *HyperClient) JsonFromCmdline(cmdArgs, cmdEnvs, cmdPortmaps []string,
...
 }
 
+func parseVolume(volStr string) (*pod.UserVolume, *pod.UserVolumeReference, error) {
+
+	var (
+		vol      = pod.UserVolume{}
+		volRef   = pod.UserVolumeReference{}
+		srcName  string
+		destPath string
+		readOnly string
```
laijs: readOnly = false.
and set it to true in the first branch below when needed.

client/run.go
```diff
@@ -290,6 +303,66 @@ func (cli *HyperClient) JsonFromCmdline(cmdArgs, cmdEnvs, cmdPortmaps []string,
 	return string(jsonString), nil
 }
 
+func parseVolume(volStr string) (*pod.UserVolume, *pod.UserVolumeReference, error) {
+
+	var (
+		vol      = pod.UserVolume{}
+		volRef   = pod.UserVolumeReference{}
```
laijs: add volDriver string

client/run.go
```diff
@@ -290,6 +303,66 @@ func (cli *HyperClient) JsonFromCmdline(cmdArgs, cmdEnvs, cmdPortmaps []string,
...
+		destPath = fields[0]
+		readOnly = "rw"
+	} else {
+		return nil, nil, fmt.Errorf("flag format should be like : --volume=[host-src:]container-dest[:rw|ro]")
+	}
+
+	vol.Source = srcName
+	_, vol.Name = path.Split(vol.Source)
+
+	vol.Driver = "vfs"
```
laijs:
```
vol := pod.UserVolume{
        Driver : volDriver,
        Source: srcName,
      ....
}
```

volRef needs to be like this also.

crook:
@laijs Thanks for your comments. and I have refined the codes. please help check

client/run.go
```diff
@@ -306,58 +306,59 @@ func (cli *HyperClient) JsonFromCmdline(cmdArgs, cmdEnvs, cmdPortmaps []string,
...
-	_, vol.Name = path.Split(vol.Source)
-
-	vol.Driver = "vfs"
-	if vol.Source == "" {
+	_, volName = path.Split(srcName)
+	if srcName == "" {
 		// Set default volume driver and use destPath as volume Name
-		vol.Driver = ""
-		_, vol.Name = path.Split(destPath)
+		volDriver = ""
```
laijs: move `volDriver = ""` up in one of the branches.

crook: what if '-v "":/hello' format? the driver is empty or VFS?

client/run.go
```diff
@@ -254,6 +258,15 @@ func (cli *HyperClient) JsonFromCmdline(cmdArgs, cmdEnvs, cmdPortmaps []string,
 		}
 	}
 
+	for _, v := range cmdVols {
+		vol, volRef, err := parseVolume(v)
+		if err != nil {
+			return "", err
+		}
+		volumes = append(volumes, *vol)
+		volumesRef = append(volumesRef, *volRef)
```
laijs: I prefer to generate the volume name here. 
fmt.Sprintf("vol-%d", i); `i` is the order of the volumes.

client/run.go
```diff
@@ -290,6 +303,67 @@ func (cli *HyperClient) JsonFromCmdline(cmdArgs, cmdEnvs, cmdPortmaps []string,
 	return string(jsonString), nil
 }
 
+func parseVolume(volStr string) (*pod.UserVolume, *pod.UserVolumeReference, error) {
```
laijs: just pod.UserVolume is Ok, don't need to be pointer

crook: The parseVolume is like parsePortMapping, I prefer to keep consistent. And pointer is faster and less resouce

client/run.go
laijs: the order doesn't matter, but it is the source for the uniq identification.

client/run.go
laijs: speed and resouce are unimportant here.
And in golang, pointer for small object is slower and uses more resource, golang always allocate heap memory for pointers.

client/run.go
crook: make sense. But it's better to put all the codes into one function

client/run.go
laijs: Good catch! it is empty. ignore my request.

client/run.go
laijs: just add a volName argument to the parseVolumes()

crook: There is already one volStr argument. User will be confused if add more volName, right?
. The vol Name will be aded 5 random number in routine parseVolume, which is enough to make the name uniq.

client/run.go
crook: make sense, do you need me also update the parsePortMapping routine? if no need. please help close the PR, thanks.

client/run.go
```diff
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"path"
```
laijs: use path/filepath

client/run.go
```diff
@@ -290,6 +303,67 @@ func (cli *HyperClient) JsonFromCmdline(cmdArgs, cmdEnvs, cmdPortmaps []string,
...
+		srcName = fields[0]
+		destPath = fields[1]
+	} else if len(fields) == 1 {
+		// -v container-dest
+		destPath = fields[0]
+	} else {
+		return nil, nil, fmt.Errorf("flag format should be like : --volume=[host-src:]container-dest[:rw|ro]")
+	}
+
+	_, volName = path.Split(srcName)
```
laijs: how about moving it to else branch ?

crook: can't catch you

client/run.go
crook: won't update this. got this puzzle if I don't use pointer (): https://www.reddit.com/r/golang/comments/2xmnvs/returning_nil_for_a_struct/

client/run.go
crook: you are right. I should use path/filepath.Join

laijs:
could you squash the commits into one please.

client/run.go
laijs: here are two "_, volName =", the value depends on the "if else" branches, could you move it into the branch?

crook: got you point. I will update

crook:
I would like to git rebase. but don't know what's the fast/correct way? Can you advise?

client/run.go
```diff
@@ -290,6 +303,68 @@ func (cli *HyperClient) JsonFromCmdline(cmdArgs, cmdEnvs, cmdPortmaps []string,
...
+		// Set default volume driver and use destPath as volume Name
+		volDriver = ""
+		_, volName = filepath.Split(destPath)
+	} else {
+		_, volName = filepath.Split(srcName)
+		// Auto create the source folder on the host , otherwise hyperd will complain
+		if _, err := os.Stat(srcName); err != nil && os.IsNotExist(err) {
+			if err := os.MkdirAll(srcName, os.FileMode(0777)); err != nil {
+				return nil, nil, err
+			}
```
laijs: please convert the source path to abspath

laijs: and please check the destPath, ensure it starts with /

crook: ok. #431

PR #431, opened 2016-10-09
Add check host-src and container-dest folder should be abspath

crook:
As title, the host-src and container-dest folder should be abspath when 'hyperctl --v'

@laijs

### DIFF
--- a/client/run.go
+++ b/client/run.go
@@ -335,11 +335,16 @@ func parseVolume(volStr string) (*pod.UserVolume, *pod.UserVolumeReference, erro
 		return nil, nil, fmt.Errorf("flag format should be like : --volume=[host-src:]container-dest[:rw|ro]")
 	}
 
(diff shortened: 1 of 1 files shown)


client/run.go
```diff
@@ -335,11 +335,18 @@ func parseVolume(volStr string) (*pod.UserVolume, *pod.UserVolumeReference, erro
...
+		return nil, nil, fmt.Errorf("The container-dir must always be an absolute path")
+	}
+
 	if srcName == "" {
 		// Set default volume driver and use destPath as volume Name
 		volDriver = ""
 		_, volName = filepath.Split(destPath)
 	} else {
+		if !strings.HasPrefix(srcName, "/") {
+			return nil, nil, fmt.Errorf("The host-dir must always be an absolute path.")
```
laijs: just convert it to abspath

crook:
@laijs updated, User Abspath

PR #433, opened 2016-10-11
disable "alsologtostderr" as default

ptptptptptpt:


hyperd.go
```diff
@@ -55,7 +55,7 @@ func main() {
 	flMirrors := flag.String("registry_mirror", "", "Prefered docker registry mirror")
 	flInsecureRegistries := flag.String("insecure_registry", "", "Enable insecure registry communication")
 	flHelp := flag.Bool("help", false, "Print help message for Hyperd daemon")
-	flag.Set("alsologtostderr", "true")
+	flag.Set("alsologtostderr", "false")
```
gnawux: I think you could remove this line simply. 

And if you set it to false, could we open it through the command line parameter?

PR #443, opened 2016-11-03
hypercli: don't setup workdir for container by default

gao-feng:
otherwise hyperd will ignore the workdir value got from
docker inspect image.


bergwolf:
LGTM!

PR #460, opened 2016-11-27
add VERSION default

jgillich:
This makes it possible to import the client from another project without running the Makefile

I'm not sure if that's the best solution, but having to run the Makefile is a major PITA when hyperd/client is a dependency. Or maybe I'm missing something since I'm just getting started with Go, but the Go creators seem to have decided that proper build systems are unnecessary.

### DIFF
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -17,7 +17,7 @@ import (
 
 var (
 	GITCOMMIT string = "0"
-	VERSION   string
+	VERSION   string = "0.7.0"
 
 	IAMSTATIC string = "true"
 	INITSHA1  string = ""


jgillich:
ping @gnawux :)

gnawux:
@jgillich The variable is defined in build command line: https://github.com/hyperhq/hyperd/blob/master/Makefile.am#L19

jgillich:
I've tried setting that, but when I run my project with `go run -ldflags "-X github.com/hyperhq/hyperd/utils.VERSION=0.7.0" main.go`, it's still empty. hyperd is installed to `vendor/` by glide, not sure if that makes a difference.

jgillich:
Aww, turns out when using the vendor directory, the path is relative from the main project.
```sh
go run -ldflags "-X github.com/my/project/vendor/github.com/hyperhq/hyperd/utils.VERSION=0.7.0" main.go
```

Works 👍

gnawux:
@jgillich Didn't tried it in vendor, maybe you should pass it with `vendor/` in the path?

PR #445, opened 2016-11-03
Refactor hyperd daemon package

gnawux:
- move all the pod management procedures from runv to hyperd
- make the structures easier to maintain

//WIP, could not merge yet.

PR #464, opened 2016-12-07
devicemapper: create device before inject files

gnawux:
only for dm, this is an issue introduced by #445 


gnawux:
I suspect there are some other issue with dm, and will push other commits here.

Dont't merge now

gnawux:
Now the dm issues found in the tests are all fixed

PR #465, opened 2016-12-07
remove superfluous logging

jgillich:
This stuff is spamming my logs badly


gnawux:
I restart the test, I don't think this patch could cause any error.

PR #471, opened 2016-12-13
Update hyperstart api

laijs:
use container/process id (instead of seqnumber) for tty-set-winsize & reporting-exit-code.


gao-feng:
LGTM

PR #476, opened 2016-12-16
Implement streams attach in hyperd

laijs:


daemon/pod/streams.go
```diff
@@ -0,0 +1,107 @@
...
+	streamConfig.stdinPipe = ioutils.NopWriteCloser(ioutil.Discard)
+}
+
+// CloseStreams ensures that the configured streams are properly closed.
+func (streamConfig *StreamConfig) CloseStreams() error {
+	var errors []string
+
+	if streamConfig.stdin != nil {
+		if err := streamConfig.stdin.Close(); err != nil {
+			errors = append(errors, fmt.Sprintf("error close stdin: %s", err))
```
gnawux: Is this ok using `%s` for err?

daemon/pod/container.go
```diff
@@ -888,6 +884,14 @@ func (c *Container) addToSandbox() error {
 		return err
 	}
 
+	c.streams.NewInputPipes()
+	tty := &hypervisor.TtyIO{
+		Stdin:  c.streams.Stdin(),
+		Stdout: c.streams.Stdout(),
+		Stderr: c.streams.Stderr(),
+	}
+	c.p.sandbox.Attach(tty, c.Id(), nil)
```
gnawux: Then every container will listen the tty stream even if one does not have log at all, right?

If yes, I like it.

PR #481, opened 2016-12-20
should init p.sandbox earlier

gnawux:
otherwise it may lead to a panic


gao-feng:
LGTM

PR #482, opened 2016-12-20
separate start and attach in the client side

gnawux:


gnawux:
In both Rest and gRPC APIs, the attach and tty info are removed from the StartPod API, the client calls Attach API before StartPod in case it is called with attach flag.

Please take a look @laijs @feiskyer

laijs:
LGTM

feiskyer:
LGTM

PR #484, opened 2016-12-22
remove duplicate timer in stopContainers

gao-feng:
WaitProcess has the same timer


laijs:
LGTM

Issue #487, opened 2016-12-23
UserPod.Validate() dose not work as expected


resouer:
**How to reproduce:**

```
hyperctl run -t -v /tmp:/test busybox /bin/sh

ERROR: volume tmpXXX does not exist in volume list
```

**Where is Wrong:**

This check will always fail because `vset` is always empty. There's no volumes set in `PodSpec`, volumes only exist in `PodSpec.containers`
```
		for _, v := range container.Volumes {
			if _, ok := vset[v.Volume]; !ok {
				return fmt.Errorf("in container %d, volume %s does not exist in volume list.", idx, v.Volume)
			}
		}
```

PR #495, opened 2016-12-28
Remove pod volume validation

resouer:
Fixes: https://github.com/hyperhq/hyperd/issues/487

Since we do not set volumes in pod anymore (correct me if I'm wrong), why don't remove the validate process?

Signed-off-by: Harry Zhang <harryz@hyper.sh>

### DIFF
--- a/types/validate.go
+++ b/types/validate.go
@@ -83,7 +83,7 @@ func (pod *UserPod) Validate() error {
 		}
 
 		for _, v := range container.Volumes {
-			if _, ok := vset[v.Volume]; !ok {
+			if _, ok := vset[v.Volume]; !ok && v.Detail == nil {
 				return fmt.Errorf("in container %d, volume %s does not exist in volume list.", idx, v.Volume)
 			}
 		}


feiskyer:
I think volume validation is still required because

1. Pod spec didn't changed, multiple containers may reference same volume.
2. A new container could reference an existing volume of the Pod, e.g. the same volume already being mounted on an existing container.
3. If the referenced volume doesn't exist, we should add it to the volume list instead of dropping it.

cc @gnawux WDYT

resouer:
cc @gnawux :/

types/validate.go
```diff
@@ -84,7 +84,8 @@ func (pod *UserPod) Validate() error {
 
 		for _, v := range container.Volumes {
 			if _, ok := vset[v.Volume]; !ok {
-				return fmt.Errorf("in container %d, volume %s does not exist in volume list.", idx, v.Volume)
+				// add container volume to pod volumes if pod volume is not provided
+				pod.Volumes = append(pod.Volumes, v.GetDetail())
```
gnawux: I think you should just pass this volume if `v.Detail` is not nil

resouer: Then should we also allow `v.Detail == nil`?

gnawux: The legacy pod specs does not carry this field, such as those for travis test.

gnawux:
retest this please, @hykins

gnawux:
retest this please, @hykins

PR #499, opened 2017-01-06
error log fix

bergwolf:
fmt %#v on error{} does not print error strings as intended:
map[string]error{"1":(*errors.errorString)(0xc82000a310), "2":(*errors.errorString)(0xc82000a330)}


gnawux:
retest this please, @hykins

PR #503, opened 2017-01-12
Remove auxiliary container support

Crazykev:
Since ipvs based service have been merged, we don't need this anymore. 

/cc @feiskyer 

### DIFF
--- a/client/api/interface.go
+++ b/client/api/interface.go
@@ -18,7 +18,7 @@ type APIInterface interface {
 
 	WinResize(id, tag string, height, width int) error
 
-	List(item, pod, vm string, aux bool) (*engine.Env, error)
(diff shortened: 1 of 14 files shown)


gnawux:
Will we have other assistant container, such as those for the pod initialization? Or will they be treat as normal container?

Crazykev:
I only find two types of container for now, service container and regular container. Hyperstart-exec container is a virtual container in hyperstart, won't affect this. I'm not sure others.

gnawux:
@feiskyer @resouer could you confirm that we (hyperd) don't need any type of assistant containers in recent months?

feiskyer:
@gnawux Yes, we don't need it any more.

feiskyer:
LGTM

PR #506, opened 2017-01-13
set workdir to user specifed value

gao-feng:
hyperctl run --workdir="/var" ...


daemon/pod/container.go
```diff
@@ -540,6 +540,10 @@ func (c *Container) describeContainer(cjson *dockertypes.ContainerJSON) (*runv.C
 		StopSignal: strings.ToUpper(cjson.Config.StopSignal),
 	}
 
+	if c.spec.Workdir != "" {
+		cdesc.Workdir = c.spec.Workdir
+	}
```
gnawux: how about make an `if...else` with the following lines

gnawux:
LGTM

PR #508, opened 2017-01-18
Fix custom container log dir do not exist

xlgao-zju:
We will not create the log dir for now, if custom container log dir do not exist.

cc @feiskyer 

### DIFF
--- a/daemon/pod/container.go
+++ b/daemon/pod/container.go
@@ -920,19 +920,22 @@ func (c *Container) initLogger() {
 	}
 
 	if c.p.factory.logCfg.Type == jsonfilelog.Name {
+		var logdir string
+
 		if c.spec.LogPath != "" {
 			ctx.LogPath = c.spec.LogPath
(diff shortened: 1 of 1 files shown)


daemon/pod/container.go
```diff
@@ -922,6 +922,14 @@ func (c *Container) initLogger() {
 	if c.p.factory.logCfg.Type == jsonfilelog.Name {
 		if c.spec.LogPath != "" {
 			ctx.LogPath = c.spec.LogPath
+			logdir := filepath.Dir(ctx.LogPath)
+
+			if err := os.MkdirAll(logdir, os.FileMode(0755)); err != nil {
```
feiskyer: dup with else block, it's better move the code after else block.

xlgao-zju: ACK

xlgao-zju:
@feiskyer updated

feiskyer:
LGTM. Thanks.

PR #509, opened 2017-01-19
fix deadlock in ReleaseAllVms()

laijs:
Foreach() holds the reader-lock for the podlist,
but p.Dissociate() reqired the writer-lock to remove the containers,
it results deadlock.

### DIFF
--- a/daemon/daemon.go
+++ b/daemon/daemon.go
@@ -330,15 +330,18 @@ func (daemon *Daemon) DestroyAndKeepVm() error {
 }
 
 func (daemon *Daemon) ReleaseAllVms() error {
-	var (
-		err error = nil
-	)
-
-	err = daemon.PodList.Foreach(func(p *pod.XPod) error {
-		return p.Dissociate()
+	var remains = []*pod.XPod{}
+	daemon.PodList.Foreach(func(p *pod.XPod) error {
+		remains = append(remains, p)
+		return nil
 	})
-
-	return err
+	for _, p := range remains {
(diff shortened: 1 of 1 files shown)


daemon/daemon.go
```diff
@@ -330,15 +330,18 @@ func (daemon *Daemon) DestroyAndKeepVm() error {
...
+	daemon.PodList.Foreach(func(p *pod.XPod) error {
+		remains = append(remains, p)
+		return nil
 	})
-
-	return err
+	for _, p := range remains {
+		glog.V(1).Infof("try to dissociate %s", p.Id())
+		if err := p.Dissociate(); err != nil {
+			glog.V(1).Infof("fail to dissociate %s: %v", p.Id(), err)
```
feiskyer: log as warning instead of info?

laijs: fixed

daemon/daemon.go
```diff
@@ -330,15 +330,18 @@ func (daemon *Daemon) DestroyAndKeepVm() error {
...
+	daemon.PodList.Foreach(func(p *pod.XPod) error {
+		remains = append(remains, p)
+		return nil
 	})
-
-	return err
+	for _, p := range remains {
+		glog.V(1).Infof("try to dissociate %s", p.Id())
+		if err := p.Dissociate(); err != nil {
+			glog.V(1).Warningf("fail to dissociate %s: %v", p.Id(), err)
```
feiskyer: `glog.Warningf`

laijs: changed again

PR #511, opened 2017-02-13
fix daemon restore

Crazykev:
1. add saveSandbox() to save runv sandbox persistence info.
2. implement some workflow in save/restore pod data.
3. fix some bugs.

For now, this PR could only support not running pod restore, others will be fixed in next PR in runv/hyperd.

### DIFF
--- a/daemon/daemon.go
+++ b/daemon/daemon.go
@@ -259,8 +259,9 @@ func (daemon *Daemon) initDefaultLog(c *apitypes.HyperConfig) {
 }
 
 func (daemon *Daemon) GetPodNum() int64 {
-	pods, err := daemon.db.ListPod()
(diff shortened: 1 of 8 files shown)


laijs:
it seems one of the pod (ID: "busybox") was failed to be removed without any notification, but the next creations of pods failed.

Crazykev:
Just notice that removing pod data from db is also not implement yet, so those test pods(including  ID:"busybox") is restored when hyperd start again. I guess this is the reason that when running  integration test the second time, many cases will fail.

laijs:
you could remove all pods at the end of every test

Crazykev:
@laijs I've tried to add removing pod/container from daemondb, it should work now.

laijs:
LGTM

PR #515, opened 2017-02-23
fix daemon restore after kill -HUP

Crazykev:
This PR enable restore running container/sandbox after kill -HUP hyperd.
Following of #511. Relay on https://github.com/hyperhq/runv/pull/441


daemon/pod/provision.go
```diff
@@ -183,7 +200,12 @@ func (p *XPod) ContainerStart(cid string) error {
 		return err
 	}
 
-	return c.start()
+	if err = c.start(); err != nil {
+		return err
+	}
+
+	// save Sandbox state for attachID changed.
+	return p.saveSandbox()
```
laijs: when Exec, attachID is also changed

PR #520, opened 2017-03-01
Fix something about stopPod

Crazykev:


daemon/pod/decommission.go
```diff
@@ -593,10 +593,14 @@ func (p *XPod) decommissionResources() (err error) {
...
 	for _, v := range p.volumes {
 		ev := v.umount()
+		v.Lock()
+		v.status = S_VOLUME_CREATED
 		if ev != nil {
+			v.status = S_VOLUME_ERROR
 			err = ev
 			v.Log(ERROR, err)
 		}
+		v.Unlock()
```
laijs: it seams better to move the code into v.umount()

Crazykev: yep, that's better. fixed.

PR #522, opened 2017-03-01
add new StorageDriver: rawblock

laijs:
```
The driver creates raw-file-block for every images/valume.
It likes vfs driver in the operation creating an image from
the parent. Both drivers do the operation by copying, but
rawblock driver tries to copy the parent by reflink.
So the copy-on-write is supported for the rawblock driver
if the backing(host) filesystem supports copy-on-write.

If the backing filesystem is btrfs, rawblock driver
supports copy-on-write now. If the backing filesystem
is xfs, the rawblock driver will test whether it is
reflink-enabled, if true, rawblock driver supports
copy-on-write.

rawblock driver vs vfs driver:
  hypervisor handles block better than dir(9p, sharefs).
  rawblock driver supports copy-on-write

rawblock driver vs devicemapper:
  loopback-based devicemapper is too ugly.
  rawblock driver (will) use loopback only when manipulate
    the filesystem in host.

The rawblock driver is also the base feature prepared for
dax support and pages-shared across VMs.

```

### DIFF
--- a/daemon/storage.go
+++ b/daemon/storage.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"math/rand"
+	"os"
 	"path/filepath"
 	"strconv"
 	"strings"
@@ -17,6 +18,7 @@ import (
 	"github.com/hyperhq/hyperd/storage"
 	"github.com/hyperhq/hyperd/storage/aufs"
 	dm "github.com/hyperhq/hyperd/storage/devicemapper"
+	"github.com/hyperhq/hyperd/storage/graphdriver/rawblock"
 	"github.com/hyperhq/hyperd/storage/overlay"
 	"github.com/hyperhq/hyperd/storage/vbox"
 	apitypes "github.com/hyperhq/hyperd/types"
@@ -42,6 +44,7 @@ var StorageDrivers map[string]func(*dockertypes.Info, *daemondb.DaemonDB) (Stora
 	"devicemapper": DMFactory,
 	"aufs":         AufsFactory,
 	"overlay":      OverlayFsFactory,
+	"rawblock":     RawBlockFactory,
 	"vbox":         VBoxStorageFactory,
 }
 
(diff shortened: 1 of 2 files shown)


laijs:
```
# hyperctl info
Images: 4
Containers: 0
PODs: 1
Storage Driver: rawblock
  Backing Filesystem: xfs
  Support Copy-On-Write: false
  Block Filesystem: xfs
  Block Size: 10GB
...
```

daemon/storage.go
```diff
@@ -400,6 +404,70 @@ func (o *OverlayFsStorage) RemoveVolume(podId string, record []byte) error {
...
+		return fmt.Errorf("Failed to create block:%v:%s", err, string(out))
+	}
+	spec.Source = block
+	spec.Format = "xfs"
+	spec.Fstype = "raw"
+	return nil
+}
+
+func (s *RawBlockStorage) RemoveVolume(podId string, record []byte) error {
+	return nil
```
gnawux: Should we remove the volume file here? Or is it created in the Pod dir?

daemon/storage.go
```diff
@@ -400,6 +404,70 @@ func (o *OverlayFsStorage) RemoveVolume(podId string, record []byte) error {
...
+
+	return vol, nil
+}
+
+func (s *RawBlockStorage) CleanupContainer(id, sharedDir string) error {
+	return nil
+}
+
+func (s *RawBlockStorage) InjectFile(src io.Reader, mountId, target, baseDir string, perm, uid, gid int) error {
+	return fmt.Errorf("unsupported")
```
gnawux: The hyperd inject resolv.conf file with `InjectFile` by default. And we have inject file cases in CI scripts. I think we need this to be implemented finally.

storage/graphdriver/rawblock/driver.go
```diff
@@ -0,0 +1,245 @@
...
+		return nil, err
+	}
+	if fsName, ok := graphdriver.FsNames[fsMagic]; ok {
+		backingFs = fsName
+	}
+
+	// check if they are running over btrfs or xfs
+	switch fsMagic {
+	case graphdriver.FsMagicBtrfs: // support
+	case graphdriver.FsMagicXfs: // TODO: check support(201609) and use ioctl(FICLONE) instead of cp --reflink
```
gnawux: This commit supports btrfs only, right?

storage/graphdriver/rawblock/driver.go
```diff
@@ -0,0 +1,245 @@
...
+	options := "loop"
+
+	if d.blockFs == "xfs" {
+		// XFS needs nouuid or it can't mount filesystems with the same fs
+		options = joinMountOptions(options, "nouuid")
+	}
+
+	options = joinMountOptions(options, label.FormatMountLabel("", mountLabel))
+
+	//if err := mount.Mount(d.block(id), d.mnt(id), d.blockFs, options); err != nil {
```
gnawux: Dose this work? or just use `exec` for debug?

storage/graphdriver/rawblock/driver.go
laijs: it supports all host filesystem. only btrfs supports cow. xfs supports cow soon.

storage/graphdriver/rawblock/driver.go
laijs: we need to add loopback related code when use mount.Mount(). using mount commandline is much simpler. The 'Put()' function forgot to detach the loopdev. it needs to be fixed.

storage/graphdriver/rawblock/driver.go
```diff
@@ -0,0 +1,245 @@
...
+		return err
+	}
+	if err := idtools.MkdirAllAs(d.mnt(id), 0755, rootUID, rootGID); err != nil {
+		return err
+	}
+	if parent == "" {
+		return d.createBaseBlock(id, mountLabel)
+	}
+	// TODO: use syscall directly, cp --reflink supports btrfs only currently
+	if out, err := exec.Command("cp", "-a", "--reflink=auto", d.block(parent), d.block(id)).CombinedOutput(); err != nil {
```
gnawux: Won't this failed if it is not on btrfs?

daemon/storage.go
laijs: the code was copied from other drivers. but it seems the volume file should be removed.

storage/graphdriver/rawblock/driver.go
laijs: "--reflink=auto", it always tries with ioctl(BTRFS_IOC_CLONE) and will fallback to normal copy(read-write) on non-btrfs. so we need to use directly syscall for xfs.

```
ioctl(4, BTRFS_IOC_CLONE, 0x3)          = -1 ENOTTY (Inappropriate ioctl for device)
fadvise64(3, 0, 0, POSIX_FADV_SEQUENTIAL) = 0
read(3, "..."..., 65536) = 252
write(4, "..."..., 252) = 252
```

gnawux: Got it. Thank you

bergwolf: BTRFS_IOC_CLONE has been pulled to generic vfs layer as FICLONE, so ioctl(BTRFS_IOC_CLONE) should work for all file systems that support file clone.

storage/graphdriver/rawblock/driver.go
```diff
@@ -0,0 +1,245 @@
...
+		return err
+	}
+	return nil
+}
+
+// Get returns the directory for the given id.
+func (d *Driver) Get(id, mountLabel string) (string, error) {
+	d.Lock()
+	defer d.Unlock()
+	active := d.active[id]
```
bergwolf: d.active does not seem reliable upon daemon restart. e.g., after daemon restarts, d.active[id] is zero and damon tries to mount the device again, which will fail if a device is mounted before.

bergwolf: Possibly we can use xattr to save the refcount in the raw block file, and retrieve it in Get()/Put() when d.active[id] is zero.

daemon/storage.go
```diff
@@ -400,6 +404,70 @@ func (o *OverlayFsStorage) RemoveVolume(podId string, record []byte) error {
...
+}
+
+func (s *RawBlockStorage) CreateVolume(podId string, spec *apitypes.UserVolume) error {
+	block := filepath.Join(s.RootPath(), "volumes", fmt.Sprintf("%s-%s", podId, spec.Name))
+	if out, err := exec.Command("truncate", fmt.Sprintf("--size=%d", storage.DEFAULT_DM_VOL_SIZE), block).CombinedOutput(); err != nil {
+		return fmt.Errorf("Failed to create block:%v:%s", err, string(out))
+	}
+	spec.Source = block
+	spec.Format = "xfs"
+	spec.Fstype = "raw"
```
laijs: rawblockDriver.blockFs can't be fetched here, we can use 'probefs()' like dm.

bergwolf:
lgtm!

gnawux:
travis build failed, looks like download lvm source failed.

gnawux:
@laijs  we need change the lvm download link to: ftp://sources.redhat.com/pub/lvm2/LVM2.2.02.131.tgz 

or newer version: 
ftp://sources.redhat.com/pub/lvm2/LVM2.2.02.168.tgz

The fedora site has retired.

PR #529, opened 2017-03-03
use -d instead of --detach-loop for detaching loop device

laijs:


laijs:
rawblock test passed

Issue #537, opened 2017-03-09
hyperd will crash under great pressure


YaoZengzeng:
I'm  now doing k8s node e2e test, the container runtime is frakti & hyperd.
However, during the process of testing, the hyperd always crashed, which makes me so annoyed.
The log is as follows:

E0309 02:24:19.847643    3965 container.go:1054] Pod[k8s_POD.0_pod-configmaps-3b91a054-046f-11e7-8d09-42010a8c0006_e2e-tests-secrets-0ss3b_3b91f88d-046f-11e7-b2fb-42010a8c0006_aa209b8e] Con[b842642bfbb9(k8s_env-test.0_pod-configmaps-3b91a054-046f-11e7-8d09-42010a8c0006_e2e-tests-secrets-0ss3b_3b91f88d-046f-11e7-b2fb-42010a8c0006_18d2fe90)] failed to umount root volume: not a directory
E0309 02:24:19.847678    3965 decommission.go:272] Pod[k8s_POD.0_pod-configmaps-3b91a054-046f-11e7-8d09-42010a8c0006_e2e-tests-secrets-0ss3b_3b91f88d-046f-11e7-b2fb-42010a8c0006_aa209b8e] Con[b842642bfbb9(k8s_env-test.0_pod-configmaps-3b91a054-046f-11e7-8d09-42010a8c0006_e2e-tests-secrets-0ss3b_3b91f88d-046f-11e7-b2fb-42010a8c0006_18d2fe90)] failed to umount root volume
E0309 02:24:19.847690    3965 container.go:64] ContainerRemove failed: not a directory
E0309 02:24:19.852968    3965 container.go:1054] Pod[k8s_POD.0_pod-update-activedeadlineseconds-3b8d6e8d-046f-11e7-afaf-42010a8c0006_e2e-tests-pods-1245p_3b8dfb4e-046f-11e7-b2fb-42010a8c0006_9acb0442] Con[adf54fc9e08e(k8s_nginx.0_pod-update-activedeadlineseconds-3b8d6e8d-046f-11e7-afaf-42010a8c0006_e2e-tests-pods-1245p_3b8dfb4e-046f-11e7-b2fb-42010a8c0006_d04ab55f)] failed to umount root volume: not a directory
[...]

PR #542, opened 2017-03-14
volume should only be removed from a running sandbox

Crazykev:

### DIFF
--- a/daemon/pod/decommission.go
+++ b/daemon/pod/decommission.go
@@ -241,6 +241,7 @@ func (p *XPod) RemoveContainer(id string) error {
 	}()
 
 	cvols := c.volumes()
+	removedVols := make([]string, 0, len(cvols))
 
 	if p.IsRunning() {
 		if c.IsRunning() {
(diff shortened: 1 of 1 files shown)


daemon/pod/decommission.go
```diff
@@ -280,31 +296,6 @@ func (p *XPod) RemoveContainer(id string) error {
...
-			if !removed {
-				v.Log(DEBUG, "volume did not unplug because it is in use")
-				continue
-			}
-			v.Log(DEBUG, "volume unplugged")
-			removedVols = append(removedVols, cv.Name)
-		}
-	}
-
-	//remove volumes from daemondb
```
feiskyer: Those are removed accidently or intended?

Crazykev: @feiskyer intendedly, I just notice remove container shouldn't remove related volume from pod(at least there is no such logic here), so this should not store to db.

feiskyer: >  remove container shouldn't remove related volume from pod

Why? I think if a volume is not referenced by any containers, it should be removed.

Crazykev: I'm not sure what the original design of this, could later created container use those volumes meant to be removed?  If not, yeah, we should remove it from pod and db.

Crazykev:
@feiskyer I've reverted changes about removing volumes from db, if it's needed, will open another PR to fix.
also /cc @gnawux do you have any comment on the disscusion?

gnawux:
Thanks @Crazykev , I don't have any more comment on the code. 

If not too complicated, could you help adding the case in #537 to the test cases of hyperd?

Crazykev:
I have no idea of why `not a directory` happens, while this one I can do something

feiskyer:
LGTM. Let's merge this first and check whether there are other potential problems by adding loads to hyperd (e.g. via kubernetes tests).

PR #543, opened 2017-03-15
Remove conflict ghost container when create container in docker engine

Crazykev:
Currently, we have both hyperd and docker engine own the container name and id registry,
I think we should relay on hyperd, since container only exist in docker engine can't be reached by hyperd.
This is a quick fix if we create container conflict with ghost container in docker engine. No matter they are left behind by:
1. not proper clean up when failed in create pod and container
2. not proper clean up when failed in restore pod and container
3. not proper user behavior (such as remove `lib` dir and leave `containers`)

### DIFF
--- a/daemon/pod/container.go
+++ b/daemon/pod/container.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


gnawux:
I think this is an issue of clean up and should fix there. What's you guys' opinion, @laijs @gao-feng @bergwolf ?

gao-feng:
It's more reasonable to fix missing cleaning up container problem, and this is not an impossible mission.

Please report the cleanup bugs when you meet them next time.

Thanks.

Crazykev:
@gnawux @gao-feng I agree there there should be a approach to fix this. since there is already an issue related, I'll close this in favor a real solution.

PR #545, opened 2017-03-15
Add regression test for #542 and #537

Crazykev:

### DIFF
--- a/hack/lib/test.sh
+++ b/hack/lib/test.sh
@@ -154,3 +154,19 @@ hyper::test::execvm() {
   test $res == aaa
   sudo hyperctl rm $id
 }
+
+# regression test for #542 and #537
+hyper::test::remove_container_with_volume() {
+  mkdir -p  ${HYPER_TEMP}/tmp
+  sed -e "s|TMPDIR|${HYPER_TEMP}/tmp|" ${HYPER_ROOT}/hack/pods/simple-volume.pod > ${HYPER_TEMP}/simple-volume.pod
+  echo "run pod with volume"
+  pod_id=$(sudo hyperctl run -d -p ${HYPER_TEMP}/simple-volume.pod | sed -ne "s/POD id is \(.*\)/\1/p")
+  echo "test pod ID is $pod_id"
+  sudo hyperctl stop $pod_id
+  echo "stop pod $pod_id"
+  echo "remove container container-with-volume in $pod_id"
+  res=$(sudo hyperctl rm -c container-with-volume > /dev/null 2>&1; echo $?)
+  sudo hyperctl rm $pod_id
+  echo "check result should be 0, got: $res"
+  test $res -eq 0
+}
--- a/hack/pods/simple-volume.pod
+++ b/hack/pods/simple-volume.pod
@@ -1,0 +1,23 @@
+{
+        "id": "test-remove-container-with-volume",
+        "containers" : [{
+            "name": "container-with-volume",
+            "image": "busybox",
+            "volumes": [{
+                "volume": "tmp",
+                "path": "/mnt",
+                "readOnly": false
+             }]
+        }],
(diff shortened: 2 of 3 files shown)


hack/lib/test.sh
```diff
@@ -154,3 +154,20 @@ hyper::test::execvm() {
   test $res == aaa
   sudo hyperctl rm $id
 }
+
+# regression test for #542 and #537
+hyper::test::remove_container_with_volume() {
+  mkdir -p  ${HYPER_TEMP}/tmp
+  sed -e "s|TMPDIR|${HYPER_TEMP}/tmp|" ${HYPER_ROOT}/hack/pods/single-volume.pod > ${HYPER_TEMP}/single-volume.pod
```
gnawux: should be `simple-volume.pod` instead of `single-volume.pod`, or you should change the file name

hack/lib/test.sh
```diff
@@ -154,3 +154,20 @@ hyper::test::execvm() {
...
+  echo "run pod with volume"
+  pod_id=$(sudo hyperctl run -d -p ${HYPER_TEMP}/single-volume.pod | sed -ne "s/POD id is \(.*\)/\1/p")
+  echo "test pod ID is $pod_id"
+  sudo hyperctl stop $pod_id
+  echo "stop pod $pod_id"
+  cid=$(sudo hyperctl list container -p $pod_id | sed -n '2p' | awk '{print $1}')
```
gnawux: I think you don't need to list the container, and could remove the container by name.

Crazykev: @gnawux just in case someone happen to change that podfile, and not update this test.

hack/lib/test.sh
```diff
@@ -154,3 +154,20 @@ hyper::test::execvm() {
...
+  pod_id=$(sudo hyperctl run -d -p ${HYPER_TEMP}/simple-volume.pod | sed -ne "s/POD id is \(.*\)/\1/p")
+  echo "test pod ID is $pod_id"
+  sudo hyperctl stop $pod_id
+  echo "stop pod $pod_id"
+  cid=$(sudo hyperctl list container -p $pod_id | sed -n '2p' | awk '{print $1}')
+  echo "remove container $cid in $pod_id"
+  res=$(sudo hyperctl rm -c $cid > /dev/null 2>&1; echo $?)
+  sudo hyperctl rm $id
+  echo "check result should be 0, got: $res"
+  test $res -eq 0
```
gnawux: you should remove the pod before the last `test $res -eq 0` to pass jenkins test.

hack/lib/test.sh
```diff
@@ -154,3 +154,20 @@ hyper::test::execvm() {
...
+  mkdir -p  ${HYPER_TEMP}/tmp
+  sed -e "s|TMPDIR|${HYPER_TEMP}/tmp|" ${HYPER_ROOT}/hack/pods/simple-volume.pod > ${HYPER_TEMP}/simple-volume.pod
+  echo "run pod with volume"
+  pod_id=$(sudo hyperctl run -d -p ${HYPER_TEMP}/simple-volume.pod | sed -ne "s/POD id is \(.*\)/\1/p")
+  echo "test pod ID is $pod_id"
+  sudo hyperctl stop $pod_id
+  echo "stop pod $pod_id"
+  cid=$(sudo hyperctl list container -p $pod_id | sed -n '2p' | awk '{print $1}')
+  echo "remove container $cid in $pod_id"
+  res=$(sudo hyperctl rm -c $cid > /dev/null 2>&1; echo $?)
```
gnawux: could you just `hyperctl rm -c container-with-volume`, then you don't need to get the container id via a list.

Crazykev:
@gnawux updated and green now.

gnawux:
@Crazykev Thank you

Issue #544, opened 2017-03-15
Can't create and start container


feiskyer:
Create container based on image `gcr.io/google_containers/k8s-dns-sidecar-amd64` failed, and no logs are found for the created container:

```
# hyperctl run -d gcr.io/google_containers/k8s-dns-sidecar-amd64:1.14.1
POD id is k8s-dns-sidecar-amd64-1.14.1-1105431549
Time to run a POD is 5406 ms

# hyperctl list container | grep k8s-dns-sidecar-amd64-1.14.1-1105431549
9e635efb753f476a2bb3fa4e3d5d9b41e24c059f8e4d1b43ee78a23472882448   k8s-dns-sidecar-amd64-1.14.1-1105431549                                                             k8s-dns-sidecar-amd64-1.14.1-1105431549

# hyperctl list | grep k8s-dns-sidecar-amd64-1.14.1-1105431549
k8s-dns-sidecar-amd64-1.14.1-1105431549                                                         k8s-dns-sidecar-amd64-1.14.1-1105431549                    
```

After starting the container again, it is shown as running, but the container process has actually down:

```
# hyperctl exec 9e635efb753f476a2bb3fa4e3d5d9b41e24c059f8e4d1b43ee78a23472882448 ps
PID   USER     TIME   COMMAND
    1 root       0:00 /init
    4 root       0:00 ps
# hyperctl exec 9e635efb753f476a2bb3fa4e3d5d9b41e24c059f8e4d1b43ee78a23472882448 ps aux
PID   USER     TIME   COMMAND
    1 root       0:00 /init
    5 root       0:00 ps aux

# hyperctl list container | grep 9e635efb753f476a2bb3fa4e3d5d9b41e24c059f8e4d1b43ee78a23472882448
9e635efb753f476a2bb3fa4e3d5d9b41e24c059f8e4d1b43ee78a23472882448   k8s-dns-sidecar-amd64-1.14.1-1105431549                                                             k8s-dns-sidecar-amd64-1.14.1-1105431549                                                         running
```

Related hyperd log:  [logs.txt](https://github.com/hyperhq/hyperd/files/843762/logs.txt)

cc @laijs @gnawux

Crazykev:
find such log fragment, seems the reason why container start failed.
```
I0315 03:14:30.698421   13728 vm_console.go:46] SB[vm-uSONjYYWGm] [CNL] try to find the user: nobody:nobody
[...]

PR #546, opened 2017-03-15
parse docker config user field properly

bergwolf:
Docker config user field supports:
[ user | user:group | uid | uid:gid | user:gid | uid:group ]

We need to translate it into runv spec.

```
[hypervsock@hyperhq]$sudo hyperctl run -d gcr.io/google_containers/k8s-dns-sidecar-amd64:1.14.1
POD id is k8s-dns-sidecar-amd64-1.14.1-9540546015
Time to run a POD is 4702 ms
[hypervsock@hyperhq]$sudo hyperctl list
POD ID                                    POD Name                                  VM name             Status
k8s-dns-sidecar-amd64-1.14.1-9540546015   k8s-dns-sidecar-amd64-1.14.1-9540546015   vm-FToockWJoz       running
[hypervsock@hyperhq]$sudo hyperctl exec k8s-dns-sidecar-amd64-1.14.1-9540546015 ps
PID   USER     TIME   COMMAND
    1 root       0:00 /init
    3 nobody     0:00 /sidecar
    7 root       0:00 ps
```

Fixes: https://github.com/hyperhq/hyperd/issues/544

### DIFF
--- a/daemon/pod/container.go
+++ b/daemon/pod/container.go
@@ -569,8 +569,19 @@ func (c *Container) describeContainer(cjson *dockertypes.ContainerJSON) (*runv.C
 	}
 
 	if cjson.Config.User != "" {
-		cdesc.UGI = &runv.UserGroupInfo{
-			User: cjson.Config.User,
+		users := strings.Split(cjson.Config.User, ":")
+		if len(users) > 2 {
+			return nil, fmt.Errorf("container %s invalid user group config: %s", cjson.Name, cjson.Config.User)
+		}
+		if len(users) == 2 {
+			cdesc.UGI = &runv.UserGroupInfo{
+				User:  users[0],
+				Group: users[1],
+			}
+		} else {
+			cdesc.UGI = &runv.UserGroupInfo{
+				User: cjson.Config.User,
+			}
 		}
 	}
 


daemon/pod/container.go
```diff
@@ -569,8 +569,13 @@ func (c *Container) describeContainer(cjson *dockertypes.ContainerJSON) (*runv.C
...
 
 	if cjson.Config.User != "" {
+		users := strings.Split(cjson.Config.User, ":")
+		if len(users) > 2 {
+			return nil, fmt.Errorf("container %s invalid user group config: %s", cjson.Name, cjson.Config.User)
+		}
 		cdesc.UGI = &runv.UserGroupInfo{
-			User: cjson.Config.User,
+			User:  users[0],
+			Group: users[1],
```
gnawux: I suspect here may lead to panic if `config.user` is something like `"root"` (no group at all)

gnawux:
I think it's better to add two cases for this, one is "nobody:nobody", and the other is "nobody"

bergwolf:
It seems we have lost user group config specified in podfile. I'll add more fix here.
```
[hypervsock@examples]$sudo hyperctl run -p user.pod
POD id is pod-YjSjNvpSAx
Time to run a POD is 4537 ms
[hypervsock@examples]$sudo hyperctl exec pod-YjSjNvpSAx ps
PID   USER     TIME   COMMAND
    1 root       0:00 /init
    3 root       0:00 sh
    4 root       0:00 ps
[hypervsock@examples]$cat user.pod
{
        "containers" : [{
            "image": "busybox",
            "user": {
                "name": "nobody",
                "group": "nobody"
            },
            "command": ["sh"]
        }]
}
```

daemon/pod/container.go
bergwolf: yup...

gnawux:
@bergwolf I think this fix is enough for #544 , you could add more fix for the pod spec in a separated PR.

Issue #548, opened 2017-03-16
user group config in podfile spec is ignored


bergwolf:
It seems we have lost user group config specified in podfile
```
[hypervsock@examples]$sudo hyperctl run -p user.pod
POD id is pod-YjSjNvpSAx
Time to run a POD is 4537 ms
[hypervsock@examples]$sudo hyperctl exec pod-YjSjNvpSAx ps
PID   USER     TIME   COMMAND
    1 root       0:00 /init
    3 root       0:00 sh
    4 root       0:00 ps
[hypervsock@examples]$cat user.pod
{
        "containers" : [{
            "image": "busybox",
            "user": {
                "name": "nobody",
                "group": "nobody"
            },
            "command": ["sh"]
        }]
}
```

gnawux:
fixed by #550

PR #550, opened 2017-03-16
respect spec user config

bergwolf:
We should respect spec user/group config and override image user/group config if spec user config is present.

Fixes https://github.com/hyperhq/hyperd/issues/548


bergwolf:
integration test timed out on aufs and overlay

gnawux:
retest this please, hykins

Issue #549, opened 2017-03-16
Kill container with tty failed


Crazykev:
How to reproduce:
```
$ sudo hyperctl run -t -d busybox sh
POD id is busybox-2151895508
Time to run a POD is 2573 ms
$ sudo hyperctl list container
Container ID                                                       Name                 POD ID               Status
e2e7649b5c8d3306160fe403a7ebd370cf45fac17a11865afc893ba619230007   busybox-2151895508   busybox-2151895508   running
$ sudo hyperctl stop -c e2e7649b5c8d3306160fe403a7ebd370cf45fac17a11865afc893ba619230007
fail to stop container e2e7649b5c8d3306160fe403a7ebd370cf45fac17a11865afc893ba619230007: Error from daemon's response: chan broken while waiting containers: map[string]bool{"e2e7649b5c8d3306160fe403a7ebd370cf45fac17a11865afc893ba619230007":true}
```

gnawux:
how about try without `-t` flag, or `kill` with signal 9

gnawux:
@Crazykev Let me explain, the stop container kill container with its `StopSignal`.

In the `busybox` image, the `StopSignal` is `SIGTERM`, which works well in non-tty mode. However, if the `sh` of `busybox` run in tty mode, `SIGTERM` will be ignored.  Instead, only SIGKILL works. More or less, I think this is an busybox issue.

A general solution is kill with `SIGKILL` if stop not takes into effect in several seconds.

Crazykev:
@gnawux Thanks, I've understood this. But I still think this is an(part of?) issue in hyperd, could we stop container just like we stop pod, if can't stop it with default(image specified) signal in graceful time, we need kill it with `SIGKILL` and return normally.
I've tried in docker, container exit with status `Exited (137)`, seems container have been killed.

gnawux:
@Crazykev Yes, I agree. Are you willing to add this mechanism? Or I will check if any other have time to make this fix now.

Crazykev:
@gnawux Glad to do this. Will do it later today or tomorrow

PR #559, opened 2017-03-19
Enable force kill container with SIGKILL if timeout

Crazykev:
ref: #549 
- I use FutureSet workflow to re-implement `stopContainers()` to enable force kill container after timeout, change the way it works, original `graceful` guarantee `stopContianers()` will return in graceful time, now ensure container have graceful time to stop with default signal before forcibly kill it.
- remove `cMap` and `eMap`, since we don't need it in current approach(Am I correct?)
- Also add timeout in `StopContainer` grpc api since CRI need it(/cc @feiskyer )


daemon/pod/decommission.go
```diff
@@ -421,96 +402,97 @@ func (p *XPod) stopAllContainers(graceful int) error {
...
-	}
-
+func (p *XPod) stopContainers(cList []string, graceful int) error {
 	p.Log(INFO, "begin stop containers %s", cList)
-	resChan := p.sandbox.WaitProcess(true, cList, graceful)
-	errChan := make(chan *containerException, 1)
 
+	future := utils.NewFutureSet()
+	waitTime := time.Duration(graceful) * time.Second
+	if graceful == 0 {
```
gnawux: IMHO, if `graceful == 0`, we should not wait, and fire force kill directly.

Crazykev: yes, if `graceful == 0`, it is force kill directly:
```
...
forceKill := graceful == 0
resChan := p.sandbox.WaitProcess(true, []string{c.Id()}, -1)
err := c.terminate(forceKill)
...
```
This `waitTime` is the time we wait for container finish event before abort this mission(or try it with SIGKILL in non-zero `graceful`)

gnawux: Then, do we need wait 5 seconds and kill again?

Crazykev: If `graceful == 0`, it just wait 5s and won't kill again(return timeout error); if kill with image specified signal(graceful != 0), it will wait graceful time to kill with SIGKILL again.
Is this 5 seconds too long?

gnawux: If `graceful == 0`, does this mean the user don't want to wait anything, just return at once?

Crazykev: Nop if we use docker:
```
$ docker stop --help

Usage:	docker stop [OPTIONS] CONTAINER [CONTAINER...]

Stop a running container.
Sending SIGTERM and then SIGKILL after a grace period

  --help             Print usage
  -t, --time=10      Seconds to wait for stop before killing it
```
or k8s way:
https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/api/v1alpha1/runtime/api.proto#L653-L658

gnawux:
@laijs could you review this patch? Its size is far beyond what I imagined before, but looks good except for the `grace==0` case.

laijs:
LGTM

PR #569, opened 2017-03-27
fix break out of for-select

Crazykev:
introduced from #559, `select` also consume a `break`, we need `return` here.
/cc @feiskyer related with test failure in https://github.com/kubernetes/frakti/pull/113


feiskyer:
LGTM

Issue #540, opened 2017-03-14
ContainerRemove failed: not a directory


feiskyer:
```
I0314 03:48:07.753316    4884 container.go:224] Pod[k8s_POD.0_nginx-2371676037-lsh60_default_8f1ddeb8-0868-11e7-be99-42010a8c0002_34e299f0] Con[2e61168ea874(k8s_nginx.0_nginx-2371676037-lsh60_default_8f1ddeb8-0868-11e7-be99-42010a8c0002_dd86986c)] retrive info &types.ContainerStatus{Name:"k8s_nginx.0_nginx-2371676037-lsh60_default_8f1ddeb8-0868-11e7-be99-42010a8c0002_dd86986c", ContainerID:"2e61168ea874907968ce11e29f0d2a2c4ca478f59ac26f1b80bcf52f68a76c1a", Phase:"failed", Waiting:(*types.WaitingStatus)(0xc4219c2a80), Running:(*types.RunningStatus)(0xc4219c2a90), Terminated:(*types.TermStatus)(0xc4229a2780)} from status &pod.ContainerStatus{State:2, CreatedAt:time.Time{sec:63625059891, nsec:828554556, loc:(*time.Location)(0x1cb5200)}, StartedAt:time.Time{sec:63625059891, nsec:931661436, loc:(*time.Location)(0x1cb5200)}, FinishedAt:time.Time{sec:63625059914, nsec:444105256, loc:(*time.Location)(0x1c6df00)}, ExitCode:0, Killed:true, RWMutex:sync.RWMutex{w:sync.Mutex{state:0, sema:0x0}, writerSem:0x0, readerSem:0x0, readerCount:1, readerWait:0}, stateChanged:(*sync.Cond)(0xc4228af7c0)}
I0314 03:48:07.754414    4884 container.go:61] ContainerRemove with request %scontainer_id:"2e61168ea874907968ce11e29f0d2a2c4ca478f59ac26f1b80bcf52f68a76c1a"
E0314 03:48:07.754477    4884 container.go:1090] Pod[k8s_POD.0_nginx-2371676037-lsh60_default_8f1ddeb8-0868-11e7-be99-42010a8c0002_34e299f0] Con[2e61168ea874(k8s_nginx.0_nginx-2371676037-lsh60_default_8f1ddeb8-0868-11e7-be99-42010a8c0002_dd86986c)] failed to umount root volume: not a directory
[...]

PR #571, opened 2017-03-28
umount root volume only if pod is running

Crazykev:
Fix another bug found in test  `ContainerRemove failed: not a directory` , ref: #540 
could be reproduced when using fs type storage driver and doing "RunPod->StopPod->RmContainer"


Crazykev:
retest this please, hykins

gnawux:
retest this please, hykins

PR #578, opened 2017-04-01
integration test: should wait before issue kill

gnawux:


gnawux:
oops, the problem should be a deadlock

PR #586, opened 2017-04-04
do not attach stderr log stream for container without tty

gnawux:
I test about 300 pods like this

```
sudo ./hyperctl run --rm -t --name "hello${i}"  busybox sh -c 'echo hello'
```

And about 4% of them lost their output because stderr closed earlier than stdin.

### DIFF
--- a/daemon/pod/container.go
+++ b/daemon/pod/container.go
@@ -1007,11 +1007,22 @@ func (c *Container) startLogging() {
 		return
 	}
 
-	stdout, stdoutStub := io.Pipe()
-	stderr, stderrStub := io.Pipe()
-	go c.AttachStreams(c.streams, false, false, c.hasTty(), nil, stdoutStub, stderrStub, nil)
+	var (
+		stdout, stderr         io.Reader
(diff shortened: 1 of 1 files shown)


gnawux:
further test shows this could not solve the issue yet.

gnawux:
img_4285

The issue is -- container stop will lead to streams breaking at once, and message will be lost if they have not been sent to the client.

gnawux:
@laijs  do you have any suggestion?

Crazykev:
This reminds another similar random test failure in frakti, which randomly missing stderr output. 
I tried with
```
sudo ./hyperctl run --rm --name "hello${i}"  busybox blabla
```
There is a chance that stderr output `exec failed: No such file or directory` is missing. Is it also related with this?
/cc @feiskyer

laijs:
LGTM. I'm going to merge it. could you create new issues if there any problem unsolved. @gnawux @Crazykev

gnawux:
@laijs Agree. This PR does a correct job, but does not solved the issue. I will file a new one and discuss with you.

PR #602, opened 2017-04-08
[RFC] use govendor instead of godep for the nested vendor issue

gnawux:
godep will introduce a nested-vendor issue when update a dep:

```
godep update github.com/hyperhq/runv/...
```

will include `github.com/hyperhq/runv/vendor/*` to vendor/ , and this is still an open issue for
 godep https://github.com/tools/godep/issues/498 and has last for more than 8 months.
Move to govendor could solve this.

However, official `go dep` tool is under development (which is not release and not fit for daily
usage yet) and plan to release with go 1.10 by the end of 2017. I am not sure whether we should
 change a vendor tool just for about 9 months. ref: https://github.com/golang/dep/wiki/Roadmap

Or is there any other solution or work around?


Crazykev:
Actually I never use this `godep update` (for not working as expected), when try to perform an update(on runv as an example) , I will 
1. remove all the entries of runv in `hyperd/Godeps/Godeps.json`
2. use `godep save` or godep-save script in `hack` to re-vendor again.

I'm not against this(don't like godep either...),  just try to offer a workaround if you don't want to change a vendor tool for 9 months...

gnawux:
It always need some manual process, which tastes bad.

Crazykev:
Indeed, then that would be invent a wheel(write a update script) vs. change a tool 😆

gnawux:
retest this please, hykins

gnawux:
retest this please, hykins

gnawux:
@laijs @gao-feng @bergwolf what's your opinion?

laijs:
Does `godep update github.com/hyperhq/runv/...` + `rm -rf vendor/github.com/hyperhq/runv/vendor/` works?

laijs:
oops, Godeps/Godeps.json can't be fix via `rm -rf vendor/github.com/hyperhq/runv/vendor/`

gnawux:
Yep, you have to update Godep.json to remove the nest vendor entries

PR #621, opened 2017-04-26
add dax support

laijs:


laijs:
updated

laijs:
updated.
It should be tested with 2.8 or newer qemu. (just use the official qemu).
And daxblock branch of the hyperstart.
And in /etc/hyper/config.

Hypervisor=qemu # or # Hypervisor=libvirt
StorageDriver=rawblock
StorageOptions=daxblock

PR #623, opened 2017-04-28
add container stopped state

bergwolf:
If a container is stopped and pod is still running, we end up hanging client on newer attach and exec. Add stopped state and deny new attach and exec when target container is stopped.

This does not address hanging client issue when attaching to a created state container, mostly because `hyperctl run` does attach before start. Do we want to change the order? I'm not convinced because otherwise client might miss some container output. What do you think? @gnawux


gnawux:
no..... the container may be restarted, and before restart, it could be attached

bergwolf:
fair enough, I'm dropping this

Issue #613, opened 2017-04-17
Should treat `name@digest` and `name:tag` format image reference as same


Crazykev:
```
$ sudo hyperctl pull busybox
Using default tag: latest
latest: Pulling from library/busybox
Digest: sha256:32f093055929dbc23dec4d03e09dfe971f5973a9ca5cf059cbfb644c206aa83f
Status: Image is up to date for busybox:latest
$ sudo hyperctl run -d busybox@sha256:32f093055929dbc23dec4d03e09dfe971f5973a9ca5cf059cbfb644c206aa83f sh
sha256:32f093055929dbc23dec4d03e09dfe971f5973a9ca5cf059cbfb644c206aa83f: Pulling from library/busybox
Digest: sha256:32f093055929dbc23dec4d03e09dfe971f5973a9ca5cf059cbfb644c206aa83f
Status: Downloaded newer image for busybox@sha256:32f093055929dbc23dec4d03e09dfe971f5973a9ca5cf059cbfb644c206aa83f
POD id is elegant_easley
Time to run a POD is 12637 ms
```

Sometimes kubernetes will relay on this feature, pull image `busybox:latest` and then create container with image `busybox@sha256:32f093055929dbc23dec4d03e09dfe971f5973a9ca5cf059cbfb644c206aa83f`.

/cc @feiskyer

gnawux:
This is an issue derived from docker 1.10.3

```
➜ docker --version
Docker version 1.10.3, build 20f81dd
➜ docker pull busybox
Using default tag: latest
latest: Pulling from library/busybox
7520415ce762: Pull complete
Digest: sha256:32f093055929dbc23dec4d03e09dfe971f5973a9ca5cf059cbfb644c206aa83f
Status: Downloaded newer image for busybox:latest
➜ docker run -itd busybox@sha256:32f093055929dbc23dec4d03e09dfe971f5973a9ca5cf059cbfb644c206aa83f
Unable to find image 'busybox@sha256:32f093055929dbc23dec4d03e09dfe971f5973a9ca5cf059cbfb644c206aa83f' locally
sha256:32f093055929dbc23dec4d03e09dfe971f5973a9ca5cf059cbfb644c206aa83f: Pulling from library/busybox
Digest: sha256:32f093055929dbc23dec4d03e09dfe971f5973a9ca5cf059cbfb644c206aa83f
[...]

PR #624, opened 2017-05-03
work around: pull: get digest for tagged pull if it has one

gnawux:
This is a work around for #613

## Background

We have docker daemon 1.10.3 as vendor, which provides the image DB management functions. However, there is an issue of it, which caused #613.

## Possible fixes

To fix this, we have three methods

- The work around in this patch
- Update to latest docker vendor
- Introduce an indirect layer for docker vendor

As the docker is changing to moby, we think we will made further update once the moby transition done. And we will apply this simple work around firstly for the 0.8.1 release.

## Known issue

Will introduce two additional (useless actually) manifest fetching, which will slow down a pull progress, esp. when you are far from the registry.

## Test result:

```
➜  sudo ./hyperctl pull busybox
Using default tag: latest
latest: Pulling from library/busybox
7520415ce762: Pull complete
Digest: sha256:32f093055929dbc23dec4d03e09dfe971f5973a9ca5cf059cbfb644c206aa83f
Status: Downloaded newer image for busybox:latest
sha256:32f093055929dbc23dec4d03e09dfe971f5973a9ca5cf059cbfb644c206aa83f: Pulling from library/busybox
Digest: sha256:32f093055929dbc23dec4d03e09dfe971f5973a9ca5cf059cbfb644c206aa83f
Status: Downloaded newer image for busybox@sha256:32f093055929dbc23dec4d03e09dfe971f5973a9ca5cf059cbfb644c206aa83f
➜  sudo ./hyperctl images
REPOSITORY          TAG                 IMAGE ID            CREATED               VIRTUAL SIZE
busybox             latest              00f017a8c2a6        2017-03-09 13:28:04   1.11 MB
➜  sudo ./hyperctl run -t busybox@sha256:32f093055929dbc23dec4d03e09dfe971f5973a9ca5cf059cbfb644c206aa83f
/ # ls
bin   dev   etc   home  lib   proc  root  sys   tmp   usr   var
/ # exit
➜
```

### DIFF
--- a/daemon/server.go
+++ b/daemon/server.go
@@ -13,6 +13,7 @@ import (
 	"github.com/golang/glog"
 	"github.com/hyperhq/hyperd/engine"
 	"github.com/hyperhq/hyperd/lib/sysinfo"
+	"github.com/hyperhq/hyperd/libmoby/distribution"
 	apitypes "github.com/hyperhq/hyperd/types"
 	"github.com/hyperhq/hyperd/utils"
 )
@@ -319,12 +320,17 @@ func (daemon *Daemon) CmdImagePull(image, tag string, authConfig *types.AuthConf
 	// compatibility.
 	image = strings.TrimSuffix(image, ":")
 
-	var ref reference.Named
+	var (
+		ref    reference.Named
+		pnamed reference.Named
+		tagged bool
+	)
 	ref, err := reference.ParseNamed(image)
 	if err != nil {
 		return err
 	}
 
+	glog.Infof("getting image: %s:%s", ref.String(), tag)
 	if tag != "" {
 		// The "tag" could actually be a digest.
 		var dgst digest.Digest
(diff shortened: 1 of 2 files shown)


daemon/server.go
```diff
@@ -319,12 +320,17 @@ func (daemon *Daemon) CmdImagePull(image, tag string, authConfig *types.AuthConf
...
+		ref    reference.Named
+		pnamed reference.Named
+		tagged bool
+	)
 	ref, err := reference.ParseNamed(image)
 	if err != nil {
 		return err
 	}
 
+	glog.Infof("getting image: %s:%s", ref.String(), tag)
```
feiskyer: It's better to use `%q` for strings in logs.

libmoby/distribution/fetch_digest.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	endpoints, repoInfo, err = lookupV2Endpoints(svc, ref)
+	if err != nil || len(endpoints) == 0 {
+		return "", err
+	}
+
+	ctx, canceller = context.WithTimeout(context.Background(), 2*time.Minute)
+	defer func() {
+		if err != nil {
```
gao-feng: seems this defer function will never be called?

gnawux: yes, never. Just got a canceller, and don't want to waste it......

Issue #601, opened 2017-04-08
CI failure case: (rawblock) device not ready during hykins test


gnawux:
only found once during test for #597. Looks unrelated with codes of the PR. record here before merge it.

Jenkins failed, init container before the device was found by the guest kernel:

```
07:53:44 I0407 07:53:43.489775    7347 vm_console.go:46] SB[vm-aPcOrtjWvO] [CNL] hyper_handle_event event EPOLLOUT, he 0x61d648, fd 3, 0x61d4c0
07:53:44 I0407 07:53:43.489781    7347 vm_console.go:46] SB[vm-aPcOrtjWvO] [CNL] hyper_modify_event modify event fd 3, 0x61d648, event 8193
07:53:44 I0407 07:53:43.491644    7347 vm_console.go:46] SB[vm-aPcOrtjWvO] [CNL] hyper_handle_event event EPOLLIN, he 0x61d648, fd 3, 0x61d4c0
07:53:44 I0407 07:53:43.491655    7347 vm_console.go:46] SB[vm-aPcOrtjWvO] [CNL] hyper_ctlfd_read: get length 581
07:53:44 I0407 07:53:43.491660    7347 vm_console.go:46] SB[vm-aPcOrtjWvO] [CNL] hyper ctl append type 14, len 4
07:53:44 I0407 07:53:43.492387    7347 vm_console.go:46] SB[vm-aPcOrtjWvO] [CNL] hyper_modify_event modify event fd 3, 0x61d648, event 8197
07:53:44 I0407 07:53:43.492396    7347 vm_console.go:46] SB[vm-aPcOrtjWvO] [CNL] hyper_ctlmsg_handle, type 17, len 581
[...]

PR #626, opened 2017-05-03
[test][do not merge] add a delay for block device

gnawux:
guest need time to initialize device ( #601 )

just for a test, don't mind I modify vendor code directly.


gnawux:
hyperhq/hyperstart#302 has fixed via add the wait code in guest

Issue #577, opened 2017-04-01
No content returned by cat command


feiskyer:
While executing a container with command `sh -c 'ls -al /all && cat /all/podname && cat /all/secret-data && cat /all/configmap-data'`, `ls` shows file size is not zero, but there is no content returned by following cat commands:

```
total 4
drwxrwxrwt    3 root     root           140 Apr  1 10:18 .
drwxr-xr-x    3 root     root          4096 Apr  1 10:18 ..
drwxr-xr-x    2 root     root           100 Apr  1 10:18 ..4984_01_04_10_18_38.047673749
lrwxrwxrwx    1 root     root            31 Apr  1 10:18 ..data -> ..4984_01_04_10_18_38.047673749
lrwxrwxrwx    1 root     root            21 Apr  1 10:18 configmap-data -> ..data/configmap-data
lrwxrwxrwx    1 root     root            14 Apr  1 10:18 podname -> ..data/podname
lrwxrwxrwx    1 root     root            18 Apr  1 10:18 secret-data -> ..data/secret-data
```

Releated hyperd log: [hyperd.log.txt](https://github.com/hyperhq/hyperd/files/887530/hyperd.log.txt)

gnawux:
could have a try with `;` instead of `&&`? with `&&`, if any of the cat return non-zero, the rest command won't be executed.

feiskyer:
@gnawux The command executes success, so there is no non-zero return code.

gnawux:
@feiskyer all the non zero size files are symlink, and can't tell if they are empty at that time.

gnawux:
any update on this issue?

feiskyer:
@laijs @gao-feng Could you help to fix the problem?

laijs:
`ls` didn't show the file size zero or not,  could you use `ls -Lal` instead please?

feiskyer:
```
# hyperctl logs dce0c7c9bd61fc7b53fa0936a0442e0b4c8b93e5000dfd06dd73014380db1938
total 16
drwxrwxrwt    3 root     root           140 Apr 13 03:31 .
drwxr-xr-x    3 root     root          4096 Apr 13 03:31 ..
drwxr-xr-x    2 root     root           100 Apr 13 03:31 ..4984_13_04_03_31_10.967674860
drwxr-xr-x    2 root     root           100 Apr 13 03:31 ..data
-rw-r--r--    1 root     root            17 Apr 13 03:31 configmap-data
[...]

PR #605, opened 2017-04-10
Do not close all streams right after container stopped

gnawux:
fix #604

The container stopping should hup all streams if there are any.
Just check and reclaim any remains after a period should be safe.

Need update #585 later if this is approved

Signed-off-by: Wang Xu <gnawux@gmail.com>

PR #627, opened 2017-05-05
Outstream

bergwolf:
The container logger copier expect to read EOF so that it can determine
the container output stream has ended. If we do not close stdout/stderr
streams, the logger copier will continue waiting for a new liner to stop
reading. When container stops and closes the logger, container logs are
lost forever.

fixes: #577

### DIFF
--- a/daemon/pod/container.go
+++ b/daemon/pod/container.go
@@ -267,8 +267,7 @@ func (c *Container) Remove() error {
 }
 
 // Container operations:
-
-func (c *Container) attach(stdin io.ReadCloser, stdout io.Writer, rsp chan<- error) error {
+func (c *Container) attach(stdin io.ReadCloser, stdout io.WriteCloser, rsp chan<- error) error {
 	if c.p.sandbox == nil || c.descript == nil {
 		err := fmt.Errorf("container not ready for attach")
 		c.Log(ERROR, err)
@@ -279,7 +278,7 @@ func (c *Container) attach(stdin io.ReadCloser, stdout io.Writer, rsp chan<- err
 	if stdout != nil {
 		if !c.hasTty() {
 			stderr = stdcopy.NewStdWriter(stdout, stdcopy.Stderr)
-			stdout = stdcopy.NewStdWriter(stdout, stdcopy.Stdout)
+			stdout = &writeCloser{stdcopy.NewStdWriter(stdout, stdcopy.Stdout), stdout}
 		}
 	}
 	detachKeys, _ := term.ToBytes(DetachKeys)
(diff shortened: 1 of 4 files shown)


daemon/pod/container.go
```diff
@@ -268,18 +268,23 @@ func (c *Container) Remove() error {
 
 // Container operations:
 
-func (c *Container) attach(stdin io.ReadCloser, stdout io.Writer, rsp chan<- error) error {
+type stdWriteCloser struct {
+	io.Writer
+	io.Closer
+}
```
gnawux: sth. similar here

https://github.com/hyperhq/hyperd/pull/605/files#diff-3a27bed0599dd83f8d78dded991998deR81

gnawux:
LGTM

gnawux:
retest this please, hykins

bergwolf:
retest this please hykins

gnawux:
retest this please, hykins

gnawux:
on the result of hykins -- http://ci.hypercontainer.io:8080/job/hyperd-auto/404/console

I am afraid that this PR will affect exec output in some case. We need make it clear before merge it.

gnawux:
Read again, looks this PR has no effect on exec...

gnawux:
I think the failure of http://ci.hypercontainer.io:8080/job/hyperd-auto/404/console is because the exec exit before the tty stream is finished. And it does not related with this PR.

daemon/pod/container.go
```diff
@@ -267,19 +267,18 @@ func (c *Container) Remove() error {
...
 	}
 
-	var stderr io.Writer
+	var stderr io.WriteCloser
 	if stdout != nil {
 		if !c.hasTty() {
-			stderr = stdcopy.NewStdWriter(stdout, stdcopy.Stderr)
-			stdout = stdcopy.NewStdWriter(stdout, stdcopy.Stdout)
+			stderr = &writeCloser{stdcopy.NewStdWriter(stdout, stdcopy.Stderr), stdout}
+			stdout = &writeCloser{stdcopy.NewStdWriter(stdout, stdcopy.Stdout), stdout}
```
gnawux: will this cause double `Close()` of `stdout`? And does this hurt?

gnawux:
test exceed 40 minutes and is killed by hykins. retest this please, hykins

daemon/pod/container.go
bergwolf: hmm, good point! Impact of a second call to Close() is undefined. We should avoid it IMO.

gnawux:
@bergwolf I think the `writeMultiCloser` is too complicated. I think just like exec does is enough --- 

- use `io.Writer` as signature instead of `io.WriteCloser`
- at the end, try cast to `io.Closer`, and `Close()` only if it is a `Closer`.
- pass `stdout` as `writeCloser`, and `stderr` as pure `Writer`

gnawux:
LGTM again

laijs:
LGTM

PR #642, opened 2017-06-08
fix aufs mount option length calculation

bergwolf:
Also add some error logs in case mount(2) fails.

The magic number `54` is a bit mysterious even the very first docker commit (https://github.com/moby/moby/commit/6d97339ca23ada27812572016ad4ff9ccffa8b09 https://github.com/moby/moby/pull/9151) did not bother to explain where it comes from...

Nevertheless, the length calculation when adding dirperm1 is clearly wrong though.


gnawux:
LGTM

PR #656, opened 2017-08-22
some update on podlist

gao-feng:


laijs:
LGTM
I think it fixes some latent bugs at least, although there is no bug referred in the commit changelog.

PR #657, opened 2017-08-24
update runv vendor

bergwolf:
runv has been refactored to meet requirements from latest containerd.
Update runv vendor to adopt these changes.


gao-feng:
need to wait iptables works assigned to @gnawux, otherwise there is no nat support for hyperd, container can not reach outside.

gnawux:
ok, I will try my best to update my PR today

gao-feng:
@bergwolf  @gnawux  any update? kvmtool and xenpv support are included in this pr.

gnawux:
The two qemu case on travis are always failed even if I close/re-open the PR, however, the failed point looks nothing related with qemu (often be pulling image failed). And the hykins test could complete successfully

bergwolf:
I've bisected the failure to runv commit https://github.com/hyperhq/runv/pull/575/commits/36881991a0ccb0cd3ab9747ee749ce80dcafaeea. And the following change would fix it (https://travis-ci.org/bergwolf/hyperd/builds/287128848):
```
diff --git a/hypervisor/qemu/qemu.go b/hypervisor/qemu/qemu.go
index 2f8ad01..b18191c 100644
--- a/hypervisor/qemu/qemu.go
+++ b/hypervisor/qemu/qemu.go
@@ -283,7 +283,7 @@ func (qc *QemuContext) AddNic(ctx *hypervisor.VmContext, host *hypervisor.HostNi
        go func() {
                // close tap file if necessary
                ev, ok := <-waitChan
-               syscall.Close(fd)
+               glog.Infof("AddNic result %v %+v", ok, ev)
                if !ok {
                        close(result)
                } else {
```

However, runv has been changed again to stop passing fd around. Yet the new runv code (after https://github.com/hyperhq/runv/pull/613) requires qemu (>2.7) to work and thus cannot run with travis ci yet.

@gnawux @gao-feng shall we require newer qemu binary or continue supporting old ones?

bergwolf:
FYI, travis only comes with trusty and xenial. And xenial stock qemu is v2.5. We'll need to provide newer version of qemu deb package to pass CI if we want to move pass v2.7.

bergwolf:
@gnawux ci passed and this is based on hyperhq/runv#616 now.

PR #660, opened 2017-08-28
make ci runnable with forked repos

bergwolf:
Add `go_import_path` to import hyperhq/hyperd.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ sudo: required
 dist: trusty
 
 language: go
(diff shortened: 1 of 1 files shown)


laijs:
How about https://docs.travis-ci.com/user/languages/go/#Go-Import-Path

bergwolf:
ah, `go_import_path` should be the official way of doing the same thing. Thanks for pointing it out!

bergwolf:
updated to use `go_import_path` instead.

Jimmy-Xu:
test this please @hykins

PR #667, opened 2017-10-04
cleanup: move src for bin files to cmd/ subdir

gnawux:
from the project root dir

Signed-off-by: Wang Xu <gnawux@gmail.com>

PR #675, opened 2017-11-01
update binary packaging rules for 1.0 build

gnawux:
- enable the debian/ubuntu deb build for ARM64 arch
- #667 changed the hyperd binary path
- hyperhq/hyperstart#341 changed the kernel path


bergwolf:
LGTM!

PR #677, opened 2017-11-15
fix storage base size support

bergwolf:
When set, it is used to set both rootfs and volume size for rawblock and devicemapper graph drivers.

e.g., having `StorageBaseSize=3gb`:

```
[root@lear hyper]# hyperctl run -v /data --rm -t busybox
/ # df -h
Filesystem                Size      Used Available Use% Mounted on
/dev/sdb                  3.0G     33.4M      3.0G   1% /
devtmpfs                 53.5M         0     53.5M   0% /dev
tmpfs                    58.1M         0     58.1M   0% /dev/shm
/dev/sda                  3.0G     32.2M      3.0G   1% /data
share_dir                 1.0M      4.0K   1020.0K   0% /etc/hosts
```

cc @minz1027

### DIFF
--- a/cmd/hyperd/hyperd.go
+++ b/cmd/hyperd/hyperd.go
@@ -111,7 +111,7 @@ func mainDaemon(opt *Options) {
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


laijs:
LGTM

Issue #679, opened 2017-11-22
Portmapping rule iptables says "Bad IP address"


m-barthelemy:
I'm using Hyperd 1.0, Debian Ubuntu package.
When creating a pod with a port mapping rule, I get:
```
Unable to setup NAT rule in HYPER chain: iptables failed: iptables --wait -t nat -I HYPER -p tcp -m tcp --dport 10000 -j DNAT --to-destination 192.168.123.8/24:10000: iptables v1.6.0: Bad IP address "192.168.123.8/24"
```

If I manually run the `iptables` command, replacing `--to-destination 192.168.123.8/24:10000` with `--to-destination 192.168.123.8:10000` (/24 removed), it works.

Is that a bug?

m-barthelemy:
Hi,
Is there anything I could do to help fixing this one? Any additional information I could provide?

resouer:
@m-barthelemy Are you using Kubernetes with hyperd my I ask? which version?

m-barthelemy:
Negative, I get that error using both `hyperctl` and Hyperd API.

gnawux:
@m-barthelemy looks like a bug, will check it.

gnawux:
@m-barthelemy did you manually configure the IP address of the container or pod?

gnawux:
@m-barthelemy I confirmed this is introduced [here](https://github.com/hyperhq/runv/commit/2e9a7596ebce8421c557fbcca15987c55c674cc7#diff-38ca951ac42fe398528410b26e3f6dbeR237) by hyperhq/runv#592 .

It will be fixed soon and add additional test for it.

gnawux:
#692 fixed the issue

@m-barthelemy if it is not fixed in your side, please reopen this issue.

PR #692, opened 2017-12-13
Fix portmapping failure: bad IP address

gnawux:
fix portmapping failure introduced by hyperhq/runv#592, in which the format of ip field was changed to cidr format.

An test case for portmapping was introduced in this PR.


gnawux:
This PR will fix #692

gnawux:
This PR will fix #679

PR #696, opened 2018-01-18
Add make option GODEBUG to generate GDB friendly binary

teawater:
The option is just available in Linux.


Jimmy-Xu:

PR #697, opened 2018-01-26
Fix restore issue

teawater:


Jimmy-Xu:

hack/lib/test.sh
```diff
@@ -262,5 +262,5 @@ END
     else
       return 1
     fi
-  done < <(sudo hyperctl run -t --rm --publish 3000:1300 busybox:latest sh -c 'echo "start" ; nc -l -p 1300')
+  done < <(sudo hyperctl run -t --rm --publish 3000:1300 busybox:latest sh -c 'echo "start" ; nc -l -p 1300 -e pwd')
```
gnawux: what's the magic 💯

daemon/pod/persist.go
```diff
@@ -410,14 +410,14 @@ func (v *Volume) removeFromDB() error {
 	return removeMessage(v.p.factory.db, fmt.Sprintf(VX_KEY_FMT, v.p.Id(), v.spec.Name), v, "volume info")
 }
 
-func (inf *Interface) saveInterface() error {
+func (inf *Interface) saveInterface(id string) error {
```
gnawux: why is the id different with the `Id` field in the data structure? should we correct the data structure instead of pass a new id here?

teawater: @gnawux the reason is inf.descript.Id is not set anywhere in hyperd.

teawater: @gnawux I made a mistake with this part.
Please see the comments of https://github.com/hyperhq/hyperd/pull/697/commits/a5ac58d322dc58ed52713dfcd3b4359d39c87348 about this.

teawater:
I really hate use master as work branch.  So close it and open other pr for the issue.

PR #702, opened 2018-02-08
hack/test-cmd.sh: change test image from "busybox" to "hyperhq/busybox"

teawater:
Meet a issue that hyper::test::portmapping hang and it is a bug of
busybox image "https://github.com/docker-library/busybox/issues/44"
(It is closed but actually it is not fixed).

After discussion with @gnawux and @bergwolf, all of us think test with a image
that under control is better than use normal busybox image.

So @bergwolf add "hyperhq/busybox" image and let test use it.


Jimmy-Xu:

PR #704, opened 2018-02-11
Fix restore issue

teawater:
https://github.com/hyperhq/hyperd/pull/697


Jimmy-Xu:

gnawux:
Read the related code again, and in short, this has not solved the issue (it's solved it in one case occasionally). I will write why and how the issue occured.

gnawux:
The issue comes from the define of interface structure. Let's come back to the struct before merging the `mtu` PR,

```
message UserInterface {
  string bridge = 1;
  string ip     = 2;
  string ifname = 3;
  string mac    = 4;
  string gateway  = 5;
  string tap    = 6;//discarded
}
```

The original definition is:

- `ifname`: the id for the device
- `tap`: device name in the host

, where the name in host could be empty and auto assigned.

**However**, this definition break the backward compatibility for hypernetes and kubernetes/frakti, which _use `ifname` for name in the host, and is optional_.

After we discard the `tap` field and use the `ifname` as the name of host device, we lose the primary key of interface data struct. 

Even with this PR, if there are more than one interfaces in the spec, there will be problem when [`initResources()`](https://github.com/hyperhq/hyperd/blob/master/daemon/pod/provision.go#L384)

```
		p.interfaces[nspec.Ifname] = inf
```

The proper solution should be

- add back a Id field in [types.proto](https://github.com/hyperhq/hyperd/blob/master/types/types.proto#L440)
   ```
     string id  = 100;
   ```
- and use the `id` in  [`initResources()`](https://github.com/hyperhq/hyperd/blob/master/daemon/pod/provision.go#L384) , check empty and assign random or indexed name for it.
- assined the `id` to description in `prepare()`, [here](https://github.com/hyperhq/hyperd/blob/master/daemon/pod/networks.go#L48) and [here](https://github.com/hyperhq/hyperd/blob/master/daemon/pod/networks.go#L60)

daemon/pod/provision.go
```diff
@@ -376,12 +376,31 @@ func (p *XPod) initResources(spec *apitypes.UserPod, allowCreate bool) error {
...
+		} else {
+			noIdInterfaces = append(noIdInterfaces, inf)
+		}
+	}
+	idx := 0
+	for _, inf := range noIdInterfaces {
+		var id string
+		for {
+			idx++
+			id = string(idx)
```
gnawux: not sure if this is what you want: https://play.golang.org/p/L6-rWGbuHbn

gnawux:
and assign the spec.id to descript.Id

PR #705, opened 2018-02-11
XPod: doStopPod: call removeFromSandbox with every p.containers

teawater:
Some resource for containers needs to be released before p.sandbox.Shutdown.

Add this patch to call removeFromSandbox with every p.containers before p.sandbox.Shutdown to release the resource.


Jimmy-Xu:

gnawux:
lgtm

PR #710, opened 2018-03-29
Fix issues around XPod.doStopPod In high-load environment

teawater:


gnawux:
what's the relation with https://github.com/hyperhq/runv/pull/673

hyperhq/runv#673 addresses the double close as well.

gnawux:
discussed with @teawater , confirmed the case

when sandbox unexpected quits right after stop command received but before stop operation beginning:

1. here https://github.com/hyperhq/hyperd/blob/master/daemon/pod/decommission.go#L370 , we set the stopping status, and do not have the `resourceLock`
2. here https://github.com/hyperhq/hyperd/blob/master/daemon/pod/decommission.go#L543 , `cleanup()` holds the lock before `doStopPod()` ( https://github.com/hyperhq/hyperd/blob/master/daemon/pod/decommission.go#L374 )
3. `cleanup()` will clean the sandbox resource and status, then set `p.sandbox` to `nil` through `decommissionResources()` at https://github.com/hyperhq/hyperd/blob/master/daemon/pod/decommission.go#L615
4. when `doStopPod()` holds `resourceLock` finally, it may access to the `nil` `p.sandbox`.

PR #711, opened 2018-04-11
XPod.Stats: fix 2 issues that will make hyperd crash

teawater:


laijs:
LGTM, p.sandbox is protected by p.resourceLock!

PR #712, opened 2018-04-11
XPod.SandboxNameWithoutLock: new function

teawater:
Some caller call XPod.SandboxName with p.statusLock.
So add XPod.SandboxNameWithoutLock.


laijs:
please use the name SandboxNameLocked()

teawater:
@laijs r u ok with the new version? :)

PR #717, opened 2018-04-27
Use snap to handle the lock issue in XPod.updatePodInfo

teawater:
https://github.com/hyperhq/hyperd/pull/708 will hurt the performance.
Change to use snap to handle the issue.


Jimmy-Xu:
@hykins retest this please

teawater:
I can make sure that status lock is used safely.

PR #720, opened 2018-05-11
Daemon.RemovePod: change p.IsAlive to !p.IsStopped

teawater:
When pod get some error, its status will be set to not running but
its qemu is still running.
So change to remove check to handle the issue.


gnawux:
retest this please, hykins

teawater:
With this patch, I got some qemu doesn't quit issue.

PR #729, opened 2018-06-19
Remove snapContainers and snapVolumes and use other way to handle the lock issue of XPod.updatePodInfo

teawater:
After taking a look at the code, I think the snap xxx is useless.
So I remove snapxxx and add lock to the code that will write p.containers and p.volumes with p.statusLock.Lock().


gnawux:
LGTM

looks reasonable. why didn't we do this before trying resource lock on info and the snap?

PR #736, opened 2018-07-10
Decommision pod resources synchronously in Remove() method

### LINKED ISSUE #733: Pods cannot launch due to no available IP addresses
To replicate: run, stop and/or remove the number of pods equal to the local address space Hyperd is using. The address space hyper uses on CentOS is `192.168.123.1/24` so 253 pods will cause this error.

After that many pods have been launched hyperd is incapable of running new pods. The error message is `hyperctl ERROR: Error from daemon's response: no available ip addresses on network`

When pods are stopped or deleted, their IPs should be freed to be used again but this does not end up happening.

The error message is coming from runv's `IPAllocator` [here](https://github.com/hyperhq/runv/blob/565496c04f13ca7e5f92ad22ddff7fa8952469dd/hypervisor/network/ipallocator/ipallocator.go#L171) when it traverses all of its ip allocation and can't find anything available which means the IPs are not being properly freed.

The IPs (and all other resources) are freed [here](https://github.com/hyperhq/hyperd/blob/b398fa40e73b3d4ee76d38a511890ab10d4c7ec9/daemon/pod/decommission.go#L557) but I believe that a race condition causes execution to never reach this point.

Right before this, [here](https://github.com/hyperhq/hyperd/blob/b398fa40e73b3d4ee76d38a511890ab10d4c7ec9/daemon/pod/decommission.go#L549) the pod `cleanup()` that's been triggered by the vm shutdown channel event checks if the pod is already stopped and quits with the idea that there is nothing to be done.

I believe this is where the error is originating. The stop or remove command usually fully wraps up shutting down the VM and changing the pod into the `STOPPED` status before this code is even run.

The synchronous shutdown code [here](https://github.com/hyperhq/hyperd/blob/b398fa40e73b3d4ee76d38a511890ab10d4c7ec9/daemon/pod/decommission.go#L51) does nothing to free the IP reserved by the pod but simply deletes everything in hyper's DB and marks the pod as stopped. 

The solution is likely to either add pod interface cleanup to the synchronous shutdown code path when the interface is deleted from leveldb, or allow the async pod cleanup to carry on with its job even if the pod is shutdown.

spearl:
fixes #733 

This will deallocate everything physically belonging to a pod right after those same things are deleted from the hyper DB.

### DIFF
--- a/daemon/pod/decommission.go
+++ b/daemon/pod/decommission.go
@@ -548,10 +548,10 @@ func (p *XPod) cleanup() {
 	defer p.resourceLock.Unlock()
 
 	p.statusLock.Lock()
-	if p.status == S_POD_STOPPED || p.status == S_POD_NONE {
+	if p.status == S_POD_STOPPED {
 		p.statusLock.Unlock()
 		return
-	} else {
+	} else if p.status != S_POD_NONE {
 		p.status = S_POD_STOPPING
 	}
 	p.statusLock.Unlock()


Jimmy-Xu:

gnawux:
ok to test, hykins

teawater:
Hi @spearl ,

I think the reason that you got the issue is after XPod.Remove timeout in https://github.com/hyperhq/hyperd/blob/60a4dd858f9a5db29e2878d67276134a93648505/daemon/pod/decommission.go#L61 , it will set the status of the pod to S_POD_NONE in https://github.com/hyperhq/hyperd/blob/60a4dd858f9a5db29e2878d67276134a93648505/daemon/pod/decommission.go#L71
Then XPod.cleanup will give up clean the pod because the status of the pod is S_POD_NONE in https://github.com/hyperhq/hyperd/blob/60a4dd858f9a5db29e2878d67276134a93648505/daemon/pod/decommission.go#L552

Even if the current patch can let XPod.Remove call decommissionResources release something.  But There is still another function p.removeSandboxFromDB() need to be called.

So what about update https://github.com/hyperhq/hyperd/blob/60a4dd858f9a5db29e2878d67276134a93648505/daemon/pod/decommission.go#L551 let XPod.cleanup can handle the S_POD_NONE but not just give it up?

Thanks,
Hui

spearl:
Hi @teawater , thanks for responding! 

`p.removeSandboxFromDB()` is actually already being called synchronously inside of the call to `p.removeFromDB()` here: https://github.com/hyperhq/hyperd/blob/60a4dd858f9a5db29e2878d67276134a93648505/daemon/pod/persist.go#L243
So the only missing call is the one to `p.decommissionResources()`.
If we remove the check for `S_POD_NONE` do we still need to check for `S_POD_STOPPED`? Would we still need to change the status to `S_POD_STOPPING` in the `else`? What purpose does this check serve? The comment explaining it above is cut off mid sentence.

Also, what is point of this check? https://github.com/hyperhq/hyperd/blob/60a4dd858f9a5db29e2878d67276134a93648505/daemon/pod/decommission.go#L574
The status will never be `S_POD_NONE` at this point since it always gets changed to `S_POD_STOPPING` above. Which means that the status will get changed from `S_POD_NONE` to `S_POD_STOPPED` if we simply remove the check above.

teawater:
Hi @spearl 

I would like to handle this issue in XPod.cleanup but not XPod.Remove because XPod.cleanup is designed to clean the resource usage when VM quit.
I think If VM doesn't quit, just release the resource usage in XPod.Remove is not a good idea.

About the code in XPod.cleanup.
I think you can remove the check of S_POD_NONE in
https://github.com/hyperhq/hyperd/blob/60a4dd858f9a5db29e2878d67276134a93648505/daemon/pod/decommission.go#L551
And update following part to change p.status to S_POD_STOPPING if p.status is not S_POD_NONE.
https://github.com/hyperhq/hyperd/blob/60a4dd858f9a5db29e2878d67276134a93648505/daemon/pod/decommission.go#L555
Then following part become useable.
https://github.com/hyperhq/hyperd/blob/60a4dd858f9a5db29e2878d67276134a93648505/daemon/pod/decommission.go#L574

Thanks,
Hui

PR #738, opened 2018-07-23
Client.Commit: Fix issue that "commit" command line with tag

teawater:
Current hyperctl commit will get error if command line has tag.
The reason is tag should be moved out as a special commit option from
repo in hyperctl.

Update Client.Commit to handle the issue.

### DIFF
--- a/client/api/commit.go
+++ b/client/api/commit.go
@@ -3,6 +3,7 @@ package api
 import (
 	"encoding/json"
 	"net/url"
+	"strings"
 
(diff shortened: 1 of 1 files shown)


teawater:
@gnawux I tried to use the docker/distribution/reference when I made the patch.  But the docker/distribution/reference that is included by hyperd doesn't have proper code to handle the issue.
There are too many private elements around it.

gnawux:
alright, merge it firstly

PR #744, opened 2018-08-30
buffer: Handle the control operations (ContainerCreate) asynchronously

teawater:
Currently, hyperd handles all the control operations synchronously.
K8S will wait the result long time if there are a lot of operations
in same time.

This patch put control operations (just support ContainerCreate now)
to buffer and put back the result at once.
The buffer will handle the result in goroutins pool (the size can be
config by BufferGoroutinesMax and BufferChannelSize).

When open buffer function, hyperd can pass most of test in hyper_test.go
but not TestRenameContainer and TestSendContainerSignal.
Because after container created, the id of container will be changed to
the id that get from docker.

### DIFF
--- a/daemon/buffer/buffer.go
+++ b/daemon/buffer/buffer.go
@@ -1,0 +1,98 @@
+package buffer
+
+import (
+	"fmt"
+	"sync"
+
+	"github.com/golang/glog"
+	"github.com/hyperhq/hyperd/daemon/pod"
+	apitypes "github.com/hyperhq/hyperd/types"
+)
+
+type Buffer struct {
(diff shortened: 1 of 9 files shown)


gnawux:
merge this after 1.1 release

gnawux:
Why just `ContainerCreate`? From my point of view, I think the `PodCreate` may suffer from the parallel requesting issue more because it leads to a sandbox creation operation.

teawater:
@gnawux 
I found ContainerCreate is the slower function than PodCreate with the high load test in the begin of this year.
I think maybe the vmtemplate and vmcache help PodCreate a lot.

